PR #13, opened 2020-06-16
updated rendering logic based on updated ibr.proto file

HTKshimo:


ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -16,49 +16,63 @@
...
+        for (const visLayer of deserializedData.visualization) {
+            coordsIndexList = visLayer.coordinate_indices;
+            coordsRangeBuffer = coordsIndexList.buffer.slice(coordsIndexList.byteOffset, coordsIndexList.byteOffset+coordsIndexList.length);
+            coordsRange = new Uint32Array(coordsRangeBuffer);
+            for (var i = 0; i < coordsRange.length; i++) {
+                coordsRange[i] = swap32(coordsRange[i]);
+            }
+            coordsRangeList.push(coordsRange);
         }
+        console.log(coordsRangeList);
```
HTKshimo, line 33: Remove console logs

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -16,49 +16,59 @@
...
 
+        // Decode Indices from data.visualization[].coordinate_indices
         var deserializedData = InternalBuildingRepresentation.read(new Pbf(data));
-        var coordsIndexList = deserializedData.visualization[0].coordinates;
-        var coordsRangeBuffer = coordsIndexList.buffer.slice(coordsIndexList.byteOffset, coordsIndexList.buffer.byteLength);
-        var coordsRange = new Uint32Array(coordsRangeBuffer);
-        for (var i = 0; i < coordsRange.length; i++) {
-            coordsRange[i] = swap32(coordsRange[i]);
+        var coordsIndexList, coordsRangeBuffer, coordsRange;
+        var coordsRangeList = [];
```
kswang2400, line 22: nit: this is probably worth encapsulating in a helper function (L22:30)

there's logic here that's specific to the proto data structure that we want to abstract away from the user. Right now the output is just a list of THREE.Line objects, but in the future we might want to give the user more access to customize what they want to show and hide.

ibr/ibr-sdk/src/renderer.js
```diff
@@ -10,49 +10,59 @@
...
-    var decodedCoordsList = [];
-    for (const coord of coords) {
-        var coordinate = coord.split(" ");
-        if (coordinate.length === 2) {
-            decodedCoordsList.push( coordinate );
-        }
+    // Decode Coordinates from data.coordinates_lookup
+    var coordsLookup = deserializedData.coordinates_lookup;
+    var coordsLookupBuffer = coordsLookup.buffer.slice(coordsLookup.byteOffset, coordsLookup.byteOffset+coordsLookup.length);
+    var coordsLookupDV = new DataView(coordsLookupBuffer);
```
kswang2400, line 30: nice! i don't see where DataView is defined though?

ibr/ibr-sdk/src/renderer.js
```diff
@@ -10,49 +10,59 @@
...
+    var coordsLookupList = [];
+    for (var i = 0; i < coordsLookup.length/4; i+=4) {
```
kswang2400, line 32: nit: 4 refers to bytes per element? let's reference [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/BYTES_PER_ELEMENT) instead of having a magic int

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -16,49 +16,59 @@
...
-        var material1 = new THREE.LineBasicMaterial( { color: 0x0000ff } );
+        // Render data into three.js objects
+        var materials = [new THREE.LineBasicMaterial( { color: 0x00ffff } ), new THREE.LineBasicMaterial( { color: 0xff0000 } )];
         var points = [], lines = [];
         var geometry, line;
-        for (const layer of layerCoordinates) {
-            for (const coordinate of layer) {
-                points.push( new THREE.Vector3( coordinate[0], coordinate[1], 0 ) );
+        for (var i = 0; i < layerCoordinates.length; i++) {
+            for (const l of layerCoordinates[i]) {
```
kswang2400, line 63: nit: let's stay away from single letter, non-index variables, it makes it hard to understand as the logic gets extended

ibr/ibr-sdk/src/renderer.js
```diff
@@ -10,49 +10,59 @@
...
+            }
+            geometry = new THREE.BufferGeometry().setFromPoints( points );
+            line = new THREE.LineLoop( geometry, materials[i] );
+            lines.push( line );
         }
-        geometry = new THREE.BufferGeometry().setFromPoints( points );
-        line = new THREE.LineLoop( geometry, material1 );
-        lines.push( line );
     }
     return lines;
```
kswang2400, line 67: so right now we have two types of lines, the boundary and the wall layer, and it isn't super obvious right now.

As this set of supported elements gets larger and larger it's going to be difficult for the reader to parse out what is what in just an array of THREE.Lines

In the next change, let's start building custom render functions for each thing so we can isolate specific elements to make the code easier to maintain and debug.

PR #15, opened 2020-06-17
Added user interactions

HTKshimo:


ibr/binToJs.js
```diff
@@ -3,48 +3,42 @@
...
-    renderer.setSize( window.innerWidth, window.innerHeight );
-    document.body.appendChild( renderer.domElement );
-    var lines = IBRSDK.renderLayer( ibrData );
-    for (const line of lines) {
-        scene.add(line);
-    }
-    camera.position.set( 0, 0, 7000 );
-    camera.lookAt( 0, 0, 0 );
-    animate();
+import { OrbitControls } from './node_modules/three/examples/jsm/controls/OrbitControls.js';
```
kswang2400, line 6: we should try to import it from the package instead of directly accessing the file path

https://stackoverflow.com/a/52447106

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -56,7 +66,11 @@
         }
 
         // Render data into three.js objects
-        var materials = [new THREE.LineBasicMaterial( { color: 0x00ffff } ), new THREE.LineBasicMaterial( { color: 0xff0000 } )];
+        var materials1 = new THREE.LineBasicMaterial( { color: 0xff0000 } );
```
kswang2400, line 69: nit: more descriptive naming

also, materials looks like just a list of duplicated linebasicsmaterial, this might be what's slowing down the renderer at scale.

can we create these on demand in L83 instead? or even better, can we create one of each and then reuse?

HTKshimo: Removed manually defined material objects and use randomly generate colors from pre-defined distinguishable selection of colors.

ibr/binToJs.js
HTKshimo: require() is not built-in for browser, only in node.js, so I don't think I can import it directly from the package in browser.

PR #16, opened 2020-06-17
Fix windows-latest filepath test errors

Ray725:
In windows-latest tests, fix

- testAddFromConfigExtraKeys
- testAddFromConfig

The tests were failing because using `os.path.join` introduces inconsistent slashes to the filepath, such as `a/b\c`.

Unfortunately, testcases which were earlier blocked by these two failing tests now have the chance to fail.

### DIFF
--- a/.github/workflows/python_tools.yml
+++ b/.github/workflows/python_tools.yml
@@ -36,7 +36,6 @@ jobs:
         python setup.py install
       working-directory: ./tools/validators/ontology_validator
     - name: Ontology Yaml Validator Generator Tests 
-      if: (matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest') 
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/python_tools.yml
```diff
@@ -36,7 +36,6 @@ jobs:
         python setup.py install
       working-directory: ./tools/validators/ontology_validator
     - name: Ontology Yaml Validator Generator Tests 
-      if: (matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest')
```
charbull, line 39: can you please delete this line. It will run all of them if you dont have an if here

Ray725:
My original fix wouldn't have worked, thanks to Ryan Mahle for pointing that out!

PR #19, opened 2020-06-22
Full Building Support Part 1 - First Level Structures parsing

HTKshimo:


ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -12,7 +12,24 @@ const TWO_POINTS = 6;
...
+     * @param {Byte} data Binary data read directly from .ibr file
+     * @return {Array.<Object>} structures List of structure objects generated from input ibr data
+     */
+    function renderStructure(data) {
+        var deserializedData = InternalBuildingRepresentation.read(new Pbf(data));
+        var structures = [];
+        if ( deserializedData.structures.length > 0 ) {
+            structures = deserializedData.structures;
+        } else {
+            structures.push( deserializedData );
```
kswang2400, line 29: this seems like it would return the whole ibr data if there's no structures, is that intended?

just a note, currently the building IBR that contains the four floors IBR has no data on it, but eventually it could/would. So this if/else would drop that data since it looks like it only does the outer layer or the inner structures but not both.

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -71,10 +139,9 @@ const TWO_POINTS = 6;
         }
 
         // Render data into three.js objects
-        var materials1 = new THREE.LineBasicMaterial( { color: 0xff0000 } );
-        var materials = [new THREE.LineBasicMaterial( { color: 0x00ffff } )];
-        for (var i = 0; i < deserializedData.visualization.length-1; i++) {
-            materials.push(materials1);
+        var materials = [];
+        for (var i = 0; i < structure.visualization.length-1; i++) {
+            materials.push(new THREE.LineBasicMaterial( { color: Colors.random() } ));
```
kswang2400, line 144: wouldn't this create a new LineBasicMaterial for each line? can we share references? e.g. if I wanted all 100 lines to be red, could I just create one LineBasicMaterial for red and reuse or do i need to create 100 LineBasicMaterial objects?

ibr/binToJs.js
```diff
@@ -23,9 +23,15 @@ function onChooseFile() {
             controls.update();
             controls.enablePan = false;
             controls.enableDamping = true;
-            var lines = IBRSDK.renderLayer( ibrData );
-            for (const line of lines) {
-                scene.add(line);
+            var structures = IBRSDK.renderStructure( ibrData );
```
kswang2400, line 26: nit: renderStructure doesn't seem to be doing any rendering, maybe unpackStructure?

HTKshimo: good point!

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: pushed top level structure to deserializedData along with possible children structures.

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: This current setup creates a color for each layer in structure.visualization. So all objects (either lineLoop or lineSegments) will be using the same material.

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -12,7 +12,24 @@ const TWO_POINTS = 6;
...
+    'use strict';
+
+    /**
+     * Decode raw ibr data into structures.
+     * @param {Byte} data Binary data read directly from .ibr file
+     * @return {Array.<Object>} structures List of structure objects generated from input ibr data
+     */
+    function unpackStructure(data) {
+        var deserializedData = InternalBuildingRepresentation.read(new Pbf(data));
+        console.log(deserializedData);
```
kswang2400, line 25: rm

PR #27, opened 2020-06-30
Multi floor

HTKshimo:
Renders each floor with different height.

### DIFF
--- a/ibr/binToJs.js
+++ b/ibr/binToJs.js
@@ -7,9 +7,6 @@
 import {OrbitControls} from
   './node_modules/three/examples/jsm/controls/OrbitControls.js';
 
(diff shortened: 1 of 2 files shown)


ibr/binToJs.js
```diff
@@ -7,6 +7,9 @@
 import {OrbitControls} from
   './node_modules/three/examples/jsm/controls/OrbitControls.js';
 
+// global variable to keep track of structure's index for height adjustment
+let structureIndex = 0;
```
kswang2400, line 11: global variables are generally bad, is there another way to implement this? I only see structureIndex referenced in extractSingleStructureData, can we keep it in there?

PR #30, opened 2020-07-01
Package code

HTKshimo:
repackage functions for easier third-party use

### DIFF
--- a/ibr/binToJs.js
+++ b/ibr/binToJs.js
@@ -19,7 +19,7 @@ import {OrbitControls} from
   checkbox.
  * @return {tag} Newly created Label tag.
  */
-function createLabel(tagName, name, parentTag, forId=undefined) {
+function createLabel(tagName, name, parentTag, forId = undefined) {
   const label = document.createElement(tagName);
   if (forId) {
     label.setAttribute('for', forId);
--- a/ibr/binToJs.js
+++ b/ibr/binToJs.js
@@ -16,9 +16,10 @@ function onChooseFile() {
   if (file) {
     const fr = new FileReader();
     fr.onload = function(evt) {
-      const scene = IBRSDK.generateScene(document.body);
-      IBRSDK.createSidebar(evt.target.result,
-          document.getElementById('layerList'), scene);
+      const bin = evt.target.result;
+      IBRSDK.renderTopIBRStructure( bin,
+          0, document.getElementById('mainCanvas') ); //top level, index 0
+      IBRSDK.createSidebar( bin, document.getElementById('layerList') );
     };
     fr.readAsArrayBuffer(file);
   }
--- a/ibr/binToJs.js
+++ b/ibr/binToJs.js
@@ -4,101 +4,6 @@
  * @author shuanglihtk@google.com (Shuang Li)
  */
 
-import {OrbitControls} from
-  './node_modules/three/examples/jsm/controls/OrbitControls.js';
-
-/**
- * Create a new HTML Label Tag based on given name string and attach it to
- * given parent tag.
- * @param {string} tagName Type of tag to be created.
- * @param {string} name InnerHTML of the Label tag.
- * @param {tag} parentTag Tag to be attached to by the newly created label
-  tag.
- * @param {string} [forId=undefined] ID to be set as the value of the newly
-  created label tag's for attribute. Needed if Label tag is created for a
-  checkbox.
- * @return {tag} Newly created Label tag.
- */
-function createLabel(tagName, name, parentTag, forId = undefined) {
-  const label = document.createElement(tagName);
-  if (forId) {
-    label.setAttribute('for', forId);
-  }
-  label.innerHTML = name;
-  parentTag.appendChild(label);
-  return label;
-}
-
-/**
- * Create checkboxes for given structure's layers and child structures.
- * @param {Object} structureData Structure to be extracted and visualized.
- * @param {String} curStructureId ID of the HTML div element to attach the
- * new structure's data.
- * @param {Object} scene object to attach the THREE.js objects
- * @param {number} thisStructureIndex the index of the structure being
- processed
- * generated from structure layer data
- */
-function extractSingleStructureData(structureData, curStructureId, scene,
-    thisStructureIndex) {
-  const curStructure = IBRSDK.unpackStructure( structureData,
-      thisStructureIndex );
-  // Create checkbox for each layer
-  if (curStructure['layers'].size !== 0) {
-    for ( const [layerName, layer] of Object.entries(curStructure['layers']) ) {
-      for ( const line of layer ) {
-        line.visible = false;
-        scene.add( line );
-      }
-      const checkBox = document.createElement('INPUT');
-      const div = document.createElement('DIV');
-      checkBox.setAttribute('type', 'checkbox');
-      checkBox.setAttribute('id', structureData.name + '_' + layerName);
-      div.appendChild(checkBox);
-      createLabel('label', layerName, div, structureData.name + '_' +
-      layerName);
-      document.getElementById(curStructureId).appendChild(div);
-      checkBox.addEventListener('change', function() {
-        if (checkBox.checked) {
-          for ( const line of layer ) {
-            line.visible = true;
-          }
-        } else {
-          for ( const line of layer ) {
-            line.visible = false;
-          }
-        }
-      });
-    }
-  }
-
-  // Create label for each child structure
-  for ( let structureIndex = 0; structureIndex <
-  curStructure['structures'].length; structureIndex++ ) {
-    const li = document.createElement('li');
-    document.getElementById(curStructureId).appendChild(li);
-    const label = createLabel('span',
-        curStructure['structures'][structureIndex].name, li);
-    label.setAttribute('class', 'arrow');
-    li.appendChild(label);
-    const ul = document.createElement('ul');
-    ul.setAttribute('class', 'nested');
-    ul.setAttribute('id', curStructure['structures'][structureIndex].name);
-    li.appendChild(ul);
-    label.addEventListener('click', function() {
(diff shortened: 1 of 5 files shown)


ibr/binToJs.js
```diff
@@ -111,52 +16,9 @@ function onChooseFile() {
...
-
-      /**
-       * Renders the scene.
-       */
-      function animate() {
-        requestAnimationFrame( animate );
-        controls.update();
-        renderer.render( scene, camera );
-      }
+      const scene = IBRSDK.generateScene(document.body);
```
kswang2400, line 19: can we be more specific than document.body? users might want to have more control over where they hook the scene into.

maybe pass in a document.getElementById('ibr-scene')?

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -22,6 +25,164 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+   * Create checkboxes for given structure's layers and child structures.
+   * @param {Object} structureData Structure to be extracted and visualized.
+   * @param {String} curStructureId ID of the HTML div element to attach the
+   * new structure's data.
+   * @param {Object} scene object to attach the THREE.js objects
+   * @param {number} thisStructureIndex the index of the structure being
+   processed
+   * generated from structure layer data
+   */
+  function extractSingleStructureData(structureData, curStructureId, scene,
```
kswang2400, line 125: good start, but i think this function can be refactored down a little bit more. right now it seems like it's doing a lot of the sidebar stuff + rendering the scene. we should decouple that and have explicit pure render functions and other data helper functions

ibr/binToJs.js
HTKshimo: used element 'mainCanvas' as parent element for all THREE.js rendering

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: separated rendering functions from sidebar creation.

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -22,21 +25,196 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
   'use strict';
 
+  let scene;
+
+  /**
+   * Create a side bar for layer and structure navigation.
+   * @param {binary} IBRRawData Raw data from IBR binary data file.
+   * @param {HTMLElement} parentElement Parent element to attach the sidebar to.
+   */
+  function createSidebar(IBRRawData, parentElement) {
```
kswang2400, line 35: variable names typically lowercase

ibr/binToJs.js
```diff
@@ -111,52 +16,10 @@ function onChooseFile() {
...
-       */
-      function animate() {
-        requestAnimationFrame( animate );
-        controls.update();
-        renderer.render( scene, camera );
-      }
+      const bin = evt.target.result;
+      IBRSDK.renderTopIBRStructure( bin,
+          0, document.getElementById('mainCanvas') ); //top level, index 0
+      IBRSDK.createSidebar( bin, document.getElementById('layerList') );
```
kswang2400, line 22: nice! this makes it a lot easier for third party developers to start using this package. two notes:

1) this encapsulates the rendering side, but we're still just passing in the binary file. This means it's very hard for developers to inspect and manipulate the data. I would suggest pulling the data into a wrapper class (e.g. ibrObject = IBRSDK.parse(bin);), then passing specific things to the render (e.g. IBRSDK.render(ibrObject.getLayer('A-WALL-E')). Plus, not including the binary parsing will make writing integration tests easier

2) Also note that `renderTopIBRStructure` implicitly forces the user to have an understanding of IBR structure, which we want to avoid. A broader and more generic method like IBRSDK.render() or renderLayers() is better

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -164,42 +344,49 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+        objects[structure.visualization[i].id].push( lineSegments );
       }
     }
     return objects;
   }
 
-  exports.renderLayer = renderLayer;
-  exports.unpackStructure = unpackStructure;
+  exports.createSidebar = createSidebar;
+  exports.generateScene = generateScene;
```
kswang2400, line 386: if `renderTopIBRStructure` and `createSidebar` are the only ones being used currently, we probably shouldn't need to export the other functions

ibr/binToJs.js
```diff
@@ -4,101 +4,6 @@
  * @author shuanglihtk@google.com (Shuang Li)
```
kswang2400, line 4: i think we can rename this file to demo.js

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -22,21 +25,196 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+
+  /**
+   * Create checkboxes for given structure's layers and child structures.
+   * @param {Object} curStructure the structure generated from
+   renderSingleIBRStructure() that will be added to sidebar.
+   * @param {String} structureName the name of the structure being processed.
+   */
+  function drawSingleStructureSidebar(curStructure, structureName) {
+    // Create checkbox for each layer
+    if (curStructure['layers'].size !== 0) {
```
kswang2400, line 127: instead of manually reading this from proto, pass in a javascript array of objects `layers`

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -22,21 +25,196 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+              line.visible = false;
+            }
+          }
+        });
+      }
+    }
+
+    // Create label for each child structure
+    for ( let structureIndex = 0; structureIndex <
+    curStructure['structures'].length; structureIndex++ ) {
```
kswang2400, line 154: instead of manually reading this from proto, pass in a javascript array of objects `structures`

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -22,21 +25,196 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+      const li = document.createElement('li');
+      document.getElementById(structureName).appendChild(li);
+      const label = createLabel('span',
+          curStructure['structures'][structureIndex].name, li);
```
kswang2400, line 158: this will break if we adjust the proto structure. we should distinguish and create a separate method for data parsing and pass in the output here

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -22,21 +25,196 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+
   /**
    * Parse top level of decoded IBR Object into structures.
-   * @param {Object} deserializedData Decoded IBR Object.
+   * @param {binary} IBRRawData Raw data from IBR binary data file.
+   * @param {number} structureIndex Index of current structure(floor).
+   * @param {HTMLElement} parentElement parent HTML element that the
+    visualization will be append on.
+   */
+  function renderTopIBRStructure(IBRRawData, structureIndex, parentElement) {
```
kswang2400, line 188: javascript variable names are typically lowercase

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -107,13 +285,14 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
+     * Converts decoded ibr structure data into three.js Line objects
+     and add them to scene.
+     * @param {Object} structure structures decoded from raw ibr data.
+     * @param {number} structureIndex overall index of the structure.
      * @return {Map.<String, List.<Object>>} objects Layer name and
-     corresponding list of three.js Line objects
+     corresponding list of three.js Line objects.
      */
-  function renderLayer(structure, structureIndex) {
+  function renderLayerAndSetToInvisible(structure, structureIndex) {
```
kswang2400, line 295: not sure what and where the set to invisible applies here? also, the return value here doesn't really match the function name (creates mapping vs rendering)

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: curStructure is parsed from the proto object in function renderSingleIBRStructure(), it is a map of {layerName: layerData, layerName: layerData...}

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: curStructure is parsed from the proto object in function renderSingleIBRStructure(), curStructure['structures'] is a list of structure objects.

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: 'SetToInvisible' is not necessary to be put on function name. I will add that to JSDoc instead.

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: curStructure is parsed from the proto object in function renderSingleIBRStructure(), curStructure['structures'] is a list of structure objects. I agree getter functions are more reliable in this case. Will add them in the next commit.

kswang2400: sounds good :)

ibr/binToJs.js
HTKshimo: 1. I will add the wrapper class and getter functions in a new PR for refactoring the data class. This PR will be for refactoring the rendering functions.

PR #34, opened 2020-07-08
implements connections message with example connectors (stairs and elevators in US-SVL-TC2)

kswang2400:
depends on #26

### DIFF
--- a/ibr/ibr.proto
+++ b/ibr/ibr.proto
@@ -36,7 +36,7 @@ message InternalBuildingRepresentation {
       // Unrestricted bi-directional access between origin and destination.
       UNRESTRICTED = 1;
 
-      // Restricted access to external structure, does not guarentee access from
+      // Restricted access to external structure, does not guarantee access from
       // destination back to origin.
       ONE_WAY = 2;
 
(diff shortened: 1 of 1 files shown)


ibr/ibr.proto
```diff
@@ -78,21 +122,25 @@ message InternalBuildingRepresentation {
...
+  // coordinate segments.
   bytes coordinates_lookup = 5;
 
   // A contained layer/visualization data (e.g. tri gross area layer,
   // external/internal walls, blocking geometry etc.)
   repeated Visualization visualization = 6;
-  
-  // Boundary of the structure
+
+  // Boundary of the structure, coordinates stored in lookup.
```
danielmi-google, line 134: I think we need to define this more.  It needs to either be the beginning and ending of the indexes in the coordinates lookup, or we need to have this be another coordinates_lookup that is only a 2D vector.

kswang2400: hm yeah, or could we just store this as it's own coordinates_lookup (x, y, x, y)?

kswang2400: @danielmi-google  can we land this and iterate when we decide if boundary needs to change?

PR #43, opened 2020-07-14
Refactor data class and add unit tests

HTKshimo:


ibr/ibr-sdk/jasmine/lib/jasmine-3.5.0/boot.js
```diff
@@ -0,0 +1,136 @@
+/**
```
kswang2400, line 1: can we add jasmine through yarn instead of adding all of its source code? `yard add jasmine` should do the trick

HTKshimo: jasmine node does not support ES6 format. Replaced jasmine testing platform with AVA to allow ES6 testing.

ibr/html/index.html
```diff
@@ -9,6 +9,8 @@
 		<script src="../node_modules/three/build/three.js"></script>
 		<script src="../node_modules/three/examples/js/utils/BufferGeometryUtils.js"></script>
 		<script type="module" src="../node_modules/ibr-sdk/build/ibr-sdk.js"></script>
+		<script type="module" src="../node_modules/ibr-sdk/class/Layer.js"></script>
+		<script type="module" src="../node_modules/ibr-sdk/class/IBRObject.js"></script>
```
kswang2400, line 13: do we have to script tag all these dependencies? Is there a way to have ibr-sdk package all of this for me so i only need to import the ibr-sdk package and not manually manage these myself?

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -6,6 +6,8 @@
 
 import {OrbitControls} from
   './../../../node_modules/three/examples/jsm/controls/OrbitControls.js';
+import {IBRObject} from
+  './../class/IBRObject.js';
```
kswang2400, line 10: we need to eventually move away from relative imports, it's not scalable as a production library

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -35,27 +37,24 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
   function createSidebar(ibrRawData, parentElement) {
     const ibrData = InternalBuildingRepresentation.read(
         new Pbf(ibrRawData));
-    // for datafiles that have top level name is ""
-    if (ibrData.name === '') {
-      ibrData.name = 'ibrData.name';
-    }
+    const curIBR = new IBRObject( ibrData );
```
kswang2400, line 40: nice! this is much cleaner as an end developer to understand :)

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -35,27 +37,24 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
     li.appendChild(ul);
     rootSpan.addEventListener('click', function() {
       rootSpan.parentElement.querySelector('.nested').classList
           .toggle('active');
       rootSpan.classList.toggle('expanded-arrow');
     });
     // root structure index 0
-    const curStructure = renderSingleIBRStructure(ibrData, 0);
-    drawSingleStructureSidebar(curStructure, ibrData.name);
+    const curStructure = renderSingleIBRStructure(curIBR, 0);
```
kswang2400, line 56: nit: curIBR looks like it might contain ALL the nested data. my suggestion would be even more specific and only pass relevant data to renderSinglIBRStructure (in a later PR)

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -285,55 +267,31 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
      and add them to scene with visibility set to false.
-     * @param {Object} structure structures decoded from raw ibr data.
+     * @param {IBRObject} curStructure IBRObject generated from current
+     structure data.
      * @param {number} structureIndex overall index of the structure.
      * @return {Map.<String, List.<Object>>} objects Layer name and
      corresponding list of three.js Line objects.
      */
-  function renderLayer(structure, structureIndex) {
+  function renderLayer(curStructure, structureIndex) {
```
kswang2400, line 276: nit: just name this as `structure` or `ibrObject`, there's no looping going on to an end user so current doesn't really add to the meaning and understanding

ibr/ibr-sdk/build/ibr-sdk.js
```diff
@@ -285,55 +267,31 @@ typeof define === 'function' && define.amd ? define(['exports'], factory) :
...
-    const coordsLookupDV = new DataView(coordsLookupBuffer);
-    const coordsLookupList = [];
-    for (let i = 0; i < coordsLookup.length; i += 4) {
-      coordsLookupList.push(coordsLookupDV.getFloat32(i, false));
-    }
-
     // Read multiple ranges from Visualization.coordinates array
     const layerCoordinates = [];
-    for (const coordsRangeItem of coordsRangeList) {
+    for (const layer of curStructure.getLayers().values()) {
```
kswang2400, line 285: we should wrap a Layer class around this as well (in the next PR)

ibr/ibr-sdk/class/Layer.js
```diff
@@ -0,0 +1,93 @@
+/**
+* Swap endianness of 32bit numbers.
+* @param {number} val 32 bit number to be swapped.
+* @return {number} 32bit number in swapped endianness.
+*/
+function swap32(val) {
```
kswang2400, line 6: nit: this is defined elsewhere already right? we should consider moving this to a shared helper file soon

ibr/ibr-sdk/class/Layer.js
```diff
@@ -0,0 +1,93 @@
...
+
+  // Decode Indices from data.visualization[].coordinate_indices
+  const coordsIndexList = layerData.coordinate_indices;
+  const coordsRangeBuffer = coordsIndexList.buffer.slice(
+      coordsIndexList.byteOffset,
+      coordsIndexList.byteOffset + coordsIndexList.length);
+  const coordsRange = new Uint32Array(coordsRangeBuffer);
+  for (let i = 0; i < coordsRange.length; i++) {
+    coordsRange[i] = swap32(coordsRange[i]);
+  }
```
kswang2400, line 30: good start! but this can be refined even more to make it even easier for another developer to access this data as coordinates. maybe consider storing this data as line segments since that will bridge even closer to how the data is rendered

ibr/ibr-sdk/class/Layer.js
```diff
@@ -0,0 +1,93 @@
...
+  },
+
+  /**
+   * Append indices of new coordinates onto the coordinates indices of the
+   layer.
+   * @param {number} coordinateIndices indices of mew coordinates to be
+   appended to the coordinates indices of the layer.
+   */
+  appendCoordinatesIndices: function( coordinateIndices ) {
+    this.coordinateIndices.append(coordinateIndices);
```
kswang2400, line 88: so coordinateIndices and coordinatesLookup are very closely coupled and it could easily get out of sync. i would suggest storing coordinates data separately and then when we do have to save back to IBR, then do the compression rather than doing surgery on the indices and lookup in real time.

ibr/test/IBRObjectTest.js
```diff
@@ -0,0 +1,104 @@
...
+});
+
+test('IBRObject does not have coordinate lookup', t => {
+  t.false(t.context.ibrObject1.hasCoordinatesLookup);
+  t.is(t.context.ibrObject1.getCoordinates(), undefined);
+});
+
+test('IBRObject has coordinate lookup', t => {
+  t.true(t.context.ibrObject2.hasCoordinatesLookup);
+  t.deepEqual(t.context.ibrObject2.getCoordinates(), [269.67669677734375, 165.02369689941406, 0, 295.67669677734375, 165.02369689941406]);
```
kswang2400, line 99: nit: this seems to have the wrong number of coordinates? (should be %3=0 for vector3d and %2=0 for vector2d right?)

ibr/ibr-sdk/class/Layer.js
kswang2400: let's take this out and land the initial work

ibr/html/index.html
HTKshimo: done

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: done

ibr/ibr-sdk/build/ibr-sdk.js
HTKshimo: wrap a Layer class? Currently each layer is a Layer object.

ibr/test/IBRObjectTest.js
HTKshimo: The example I gave has 20 elements in int8Array. Then after converting to Float32Array there should be 5 elements in the array.

kswang2400: this is Vector3D right? how do we get two 3D coordinates from 5 elements?

PR #47, opened 2020-07-16
Create a PreSubmit validation for the ontology type extensions

charbull:
Each extension/pull request for the ontology type validation need to trigger this workflow to make sure the ontology extensions are valid.


tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -595,6 +595,30 @@ def _PrintType(ns, et):
     print('  HAS REDUNDANT LOCAL FIELDS')
 
 
+def _printFindings(findings, filter_text):
+  """Prints the findings of the ontology validation
+
+  Args:
+    findings: a list of Finding objects.
+    filter_text: command line arguments. The only available argument is
+      match:<value>' which will simply perform a simple string 'contains' on
```
Ray725, line 604: Possible nit? Is there supposed to be an apostrophe after `match:<value>`? I feel like that perhaps implies regex but that's not being used here

.github/workflows/OntologyTypeValidator.yml
```diff
@@ -0,0 +1,39 @@
...
+    paths: 
+      - 'ontology/yaml/resources/**'
+
+jobs:
+  build:
+  
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
```
berkoben, line 22: no windows?

charbull: no yet, @Ray725 is working the bug for windows

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
charbull: good catch there is a missing `

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -560,8 +560,8 @@ def RunPresubmit(unmodified, modified_base, modified_client):
 
   """
 
-  findings, _ = _ValidateConfigInner(unmodified, modified_base, modified_client)
-
+  findings, _ = _ValidateConfigInner(unmodified, modified_base, modified_client, False)
+  _printFindings(findings, '')
```
berkoben, line 564: printing findings in presubmit mode is going to make a big mess because the findings will end up being printed 3x (they're already printed twice with internal presubmit).  This needs to only happen on external presubmit.  Suggest pulling the print statement into a higher level method.

charbull: I see, will leave it in RunInteractive as it was before and taking out the one for Presubmit

charbull:
@berkoben @Ray725 PTAL

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -560,8 +560,7 @@ def RunPresubmit(unmodified, modified_base, modified_client):
 
   """
 
-  findings, _ = _ValidateConfigInner(unmodified, modified_base, modified_client)
-
+  findings, _ = _ValidateConfigInner(unmodified, modified_base, modified_client, False)
```
Ray725, line 563: Out of curiosity, is it better style to do `interactive=False` for the last argument? But on line `639` it just inputs the boolean so maybe this is the convention of the codebase.

PR #49, opened 2020-07-17
Revert "Lighting entity types"

charbull:
Reverts google/digitalbuildings#44


pisuke:
I understand there are more things to be discussed, so I've left some more comments [here[(https://github.com/google/digitalbuildings/pull/44) for @tasodorff 
I think we should close this PR to revert and I can address changes with a later PR.

PR #59, opened 2020-07-22
Add better logging, no more confusing stacktrace

Ray725:


tools/validators/instance_validator/main.py
```diff
@@ -40,11 +41,17 @@
...
-  print('\nValidator starting ...')
+  print('\nValidator starting ...\n')
   filename = arg.filename
 
   # throws errors for syntax
-  parsed = dict(instance_parser.parse_yaml(filename))
+  raw_parse = instance_parser.parse_yaml(filename)
+
+  if raw_parse is None:
+    print('\nParsing error, logging provided above.')
```
charbull, line 51: "Syntax Error" instead?

Ray725: Done

PR #70, opened 2020-07-24
Create node.js.yml

kswang2400:
create javascript ci tool that lints, builds, and tests the ibr package before landing to master


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,29 @@
...
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
```
charbull: you can run it only your directory check the `working_directory`

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,29 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
```
charbull, line 7: same here and on pull request you can target only your folder

kswang2400: nice, ty ty

.github/workflows/node.js.yml
kswang2400: nice ty ty

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+
+    - name: Run Google es-linter
+      run: npm run lint
```
charbull: you can use the google js styleguide from here: https://github.com/google/styleguide

I got the pylint from here: https://github.com/google/styleguide/blob/ea180f7259da4ae90ae086c1eefbeb5f440dbea2/pylintrc

kswang2400: yeah, trying to use "eslint-config-google": "^0.14.0", from ibr package but there's some errors

.github/workflows/node.js.yml
```diff
@@ -0,0 +1,38 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
```
charbull, line 8: awesome ! you can specify here the working directory, when the push or pr matches the path then the action is triggered, do you want to add the part to match the file path?
I need to add it as well on the other actions

PR #78, opened 2020-07-29
Group layers under IBR UI

HTKshimo:
Group Visualizations and Spaces to be under their own section under each IBR.


ibr/ibr-sdk/src/constants.js
```diff
@@ -7,4 +7,13 @@ export const ONE_POINT = 3;
 export const TWO_POINTS = 6;
 
 // separate floor's z-coordinate by 100 unit length
-export const FLOOR_HEIGHT = 300;
\ No newline at end of file
+export const FLOOR_HEIGHT = 300;
+
+// structural types
+export const UNKNOWN = 0;
```
kswang2400, line 13: we should reference the enum in ibr.proto instead of creating a copy of them here

HTKshimo: done

PR #86, opened 2020-08-06
Setup PNG Parser to create new IBR file from bitmap floor plan images

HTKshimo:


ibr/png-parser/CreateIBR.py
```diff
@@ -0,0 +1,34 @@
+import ibr_pb2
+import sys
+from boundary_extraction.boundary_extraction import extract_boundary
+from imagedata_creation.imagedata_creation import get_image_data
+from utils import *
+
```
kswang2400, line 6: let's wrap this in a ```if __name__ == '__main__':``` clause

https://stackoverflow.com/a/419185

ibr/png-parser/BUILD
```diff
@@ -0,0 +1,11 @@
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
+    name = 'CreateIBR',
+    main = 'CreateIBR.py',
+    srcs = ['CreateIBR.py', utils],
```
kswang2400, line 6: how come it's utils and not utils.py?

ibr/png-parser/CreateIBR.py
```diff
@@ -0,0 +1,34 @@
+import ibr_pb2
+import sys
+from boundary_extraction.boundary_extraction import extract_boundary
+from imagedata_creation.imagedata_creation import get_image_data
+from utils import *
```
kswang2400, line 5: let's avoid `import *`, either `import utils` and then reference methods like `utils.set_visualization_image_data` or explicitly import each method 'from utils import set_visualization_image_data'

ibr/png-parser/imagedata_creation/imagedata_creation.py
```diff
@@ -0,0 +1,19 @@
...
+def get_image_data(filename):
+
+    # Read file input
+    img = cv2.imread(filename)
+    print(img.dtype)
+    print(len(img))
+    print(len(img[0]))
+    print(img.shape[1])
+    print(img.shape[0])
+
```
kswang2400, line 12: nit: delete prints

ibr/png-parser/BUILD
```diff
@@ -0,0 +1,11 @@
+load("@rules_python//python:defs.bzl", "py_binary")
+
+py_binary(
+    name = 'CreateIBR',
```
kswang2400, line 4: nit: s/CreateIBR.py /create_ibr.py/g, also build target should probably match

pep8 module/file names should be lower/snakecase 

https://www.python.org/dev/peps/pep-0008/#package-and-module-names

ps: we should probably set up a python/bazel linter (separate diff) while we're at it

ibr/png-parser/CreateIBR.py
kswang2400, line 5: nit: good python style is to separate imports into 1) standard libraries, 2) third party libraries, and 3) local modules

https://stackoverflow.com/a/20763446

i.e.
```
import sys

import ibr_pb2

from boundary_extraction.boundary_extraction import extract_boundary
from imagedata_creation.imagedata_creation import get_image_data
from utils import *
```

ibr/png-parser/REDAME.md
```diff
@@ -0,0 +1,16 @@
...
+
+1. Compile the proto definition to python classes:
+	- Install latest release of protocol buffers
+	    - go to [https://developers.google.com/protocol-buffers/docs/downloads]()
+	    - click on latest version release page
+	    - click on protobuf-$VERSION-$PLATFORM.zip to download the protocol buffers pre-built binary
+	    - unzip downloaded file
+	    - place bin/protoc somewhere in your $PATH
+	- move to current directory `cd digitalbuildings/ibr/png-parser`
+	- compile the proto definition: `protoc -I=../ibr-sdk/proto/ --python_out=. ibr.proto`
```
kswang2400, line 11: this is where the python proto is being created, but im not seeing where ibr_pb2 is explicitly defined and imported into the python module? could we be more explicit with this?

ibr/png-parser/REDAME.md
```diff
@@ -0,0 +1,16 @@
+## Instructions for Running the PNG Parser:
+
+1. Compile the proto definition to python classes:
+	- Install latest release of protocol buffers
+	    - go to [https://developers.google.com/protocol-buffers/docs/downloads]()
+	    - click on latest version release page
+	    - click on protobuf-$VERSION-$PLATFORM.zip to download the protocol buffers pre-built binary
+	    - unzip downloaded file
+	    - place bin/protoc somewhere in your $PATH
```
kswang2400, line 9: could we use a pip to manage this instead of manually downloading source?

ibr/png-parser/BUILD
HTKshimo: will punt bazel python linter to another PR.

ibr/png-parser/REDAME.md
HTKshimo: According to the protobuf official website, there are two parts of protobuf that need to be installed: Protocol Compiler and language specific Runtime library. Here we used pip to install the runtime library, but the simplest way to install the protocol compiler is to download a pre-built binary from the release page.

ibr/png-parser/REDAME.md
HTKshimo: ibr_pb2 is the default output name when compiling ibr.proto. It is imported into the create_ibr.py module. I'll add a note specifying the default output name of this command.

ibr/png-parser/REDAME.md
kswang2400: sg, thanks for looking into this :)

ibr/png-parser/BUILD
kswang2400: sg

PR #88, opened 2020-08-07
Png parser UI update

HTKshimo:
Added image_data parsing to IBR UI to display images as part of visualization layer


ibr/png-parser/create_ibr.py
```diff
@@ -1,4 +1,6 @@
 import sys
+import os
+from PIL import Image
```
kswang2400, line 3: this isn't a standard library, also, do we need to include this python package in build/requirements.txt?

ibr/png-parser/imagedata_creation/BUILD
```diff
@@ -5,5 +5,7 @@ package(default_visibility = ["//visibility:public"])
 py_library(
     name = 'imagedata_creation',
     srcs = ['imagedata_creation.py'],
-    deps = [requirement("numpy")],
+    deps = [
+        requirement("opencv-python"),
```
kswang2400, line 9: still not sure why these deps differ from requirements.txt?

PR #91, opened 2020-08-07
Png parser output setting

HTKshimo:
added name and structural type to created IBR

### DIFF
--- a/ibr/png-parser/create_ibr.py
+++ b/ibr/png-parser/create_ibr.py
@@ -10,7 +10,7 @@ import utils
 
 if __name__ == '__main__':
 
-    # Read image filename
+    # Read image filepath
     if len(sys.argv) != 2:
         print('Usage:', sys.argv[0], 'image file name')
         sys.exit(-1)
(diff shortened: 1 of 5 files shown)


ibr/png-parser/create_ibr.py
```diff
@@ -10,27 +10,31 @@
...
 
     # Initialize new IBR
     new_ibr = ibr_pb2.InternalBuildingRepresentation()
     vis = ibr_pb2.InternalBuildingRepresentation.Visualization()
-    filename = sys.argv[1]
+    filepath = sys.argv[1]
+    filename = filepath.split('/')[-1]
+    filename_no_extension = ''.join(filename.split('.')[:-1])
+    utils.set_IBR_name(new_ibr, filename_no_extension)
+    utils.set_IBR_structural_type(new_ibr, ibr_pb2.InternalBuildingRepresentation.StructuralType.SPACE)
```
kswang2400, line 25: leave a TODO here since this is temporary

ibr/png-parser/create_ibr.py
```diff
@@ -43,7 +47,7 @@
     new_ibr.visualization.append(vis)
```
kswang2400, line 47: we should probably create wrapper classes for these instead of modifying raw IBR protos

PR #109, opened 2020-08-18
Add temporary connections_universe functionality

Ray725:
Connections are not currently being generated in the ontology universe, so the instance validator cannot validate the contents of a given building configuration's `connections` key. To make this eventual validation easier to implement, this PR mocks the existence and use of a `connections_universe`.

Depending on how connections end up actually being generated, slight modifications of this will be needed.


tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -42,6 +42,32 @@ def __init__(self, entity, universe, config_entity_names):
...
+      return True
+
+    if self.connections not in self.entity.keys():
+      # connections are not included in this building config
+      return True
+
+    connections_map = dict(self.entity[self.connections])
+    for device_name in connections_map.keys():
+      connection = connections_map[device_name]
+      if connection not in self.universe.connections_universe:
```
berkoben, line 66: nit: you can use set.difference() to do this in one line.

tools/validators/instance_validator/tests/entity_instance_test.py
```diff
@@ -288,5 +290,37 @@ def testValidateLinks(self):
     if not instance.IsValidEntityInstance():
       self.fail('exception incorrectly raised')
 
+  def testGoodConnections(self):
```
berkoben, line 293: this test could be stricter:  To ensure that this is actually testing what it says it is, the input should be checked to confirm that the entity instance has connections in it and that the connections universe is not null

tools/validators/instance_validator/tests/entity_instance_test.py
```diff
@@ -288,5 +290,37 @@ def testValidateLinks(self):
...
+                     'bad_building_connections.yaml'))
+    parsed = dict(parsed)
+    entity_name = list(parsed.keys())[0]
+
+    entity = dict(parsed[entity_name])
+    instance = entity_instance.EntityInstance(entity,
+                                              self.universe,
+                                              parsed.keys())
+
+    if instance.IsValidEntityInstance():
```
berkoben, line 322: shouldn't we also check that the correct finding is generated? (I assume we're doing findings?)

tools/validators/instance_validator/tests/entity_instance_test.py
```diff
@@ -288,5 +290,37 @@ def testValidateLinks(self):
...
+    entity_name = list(parsed.keys())[0]
+
+    entity = dict(parsed[entity_name])
+    instance = entity_instance.EntityInstance(entity,
+                                              self.universe,
+                                              parsed.keys())
+
+    if not instance.IsValidEntityInstance():
+      self.fail('exception incorrectly raised')
+
```
berkoben, line 308: TODO:  Add tests for the more advanced syntax

tools/validators/instance_validator/validate/entity_instance.py
Ray725: Done

tools/validators/instance_validator/tests/entity_instance_test.py
Ray725: Done

tools/validators/instance_validator/tests/entity_instance_test.py
Ray725: The instance validator isn't using findings right now. The check is less strict, and we focus on whether the console output is correct, because that's what clients will be using the validator for.

tools/validators/instance_validator/tests/entity_instance_test.py
Ray725: The syntax tests for connections already exist in a separate folder, and you can see them here: https://github.com/google/digitalbuildings/blob/master/tools/validators/instance_validator/tests/instance_parser_test.py.

berkoben: sorry, by syntax I meant the more advanced configuration.

PR #115, opened 2020-09-04
Add validation for fields on the translation based on the entity type

charbull:


tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -143,32 +143,50 @@ def _ValidateTranslation(self):
...
+    entity_type = type_parse[1]
+    entity_type = self.universe.GetEntityType(namespace, entity_type)
+    all_fields_dict = entity_type.GetAllFields()
+
     # iterate through each translation device key and determine its form
-    for device_name in translation_body.keys():
+    for field_name in translation_body.keys():
       # check if keys are UDMI compliant
-      if isinstance(translation_body[device_name].data, str):
+      if isinstance(translation_body[field_name].data, str):
```
berkoben, line 158: Add comment why this case is skipped.  It's not immediately clear to my we wouldn't check every key.

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -143,32 +143,50 @@ def _ValidateTranslation(self):
...
         continue
 
-      device_map = translation_body[device_name]
+      #check if the field_name is on the type
+      # TODO(charbull), the key in the dictionary all_fields_dict starts with `/`, needs to be cleaned
+      # pop the field out
+      key_field_name = '/'+field_name.data
+      opt_wrapper_field = all_fields_dict.pop(key_field_name, None)
```
berkoben: This is also error prone as written in the case of types that don't use global fields because it's not handling namespace.  

Recall the rules "" ed is config text: 
"FIELD"  could be LOCAL/FIELD or /FIELD depending on whether that field is defined locally in the namespace of the type
"OTHER/FIELD" is always OTHER/FIELD

so the field from the config should be split on / and the parts compared.  For unqualified fields, you need to check /FIELD and LOCAL/FIELD.  

Bonus points for not doing string manipulation other than splitting the value in the config.

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -143,32 +143,50 @@ def _ValidateTranslation(self):
...
+
+      translation_map = translation_body[field_name]
       valid_units = self.universe.unit_universe.GetUnitsMap('').keys()
       valid_states = self.universe.state_universe.GetStatesMap('').keys()
 
       # three remaining possibilities for translation format
-      if self.unit_values in device_map.keys():
-        unit_values_map = device_map[self.unit_values]
+      if self.unit_values in translation_map.keys():
+        unit_values_map = translation_map[self.unit_values]
```
berkoben, line 177: this can be stricter because the units are limited by the measurement type of the field.

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -143,32 +143,50 @@ def _ValidateTranslation(self):
...
         for unit in unit_values_map.keys():
           if unit not in valid_units:
             print('Invalid translation unit', unit)
             return False
-      elif self.states in device_map.keys():
-        states_map = device_map[self.states]
+      elif self.states in translation_map.keys():
+        states_map = translation_map[self.states]
```
berkoben, line 183: this can be stricter because states are dictated by field.

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -143,32 +143,50 @@ def _ValidateTranslation(self):
...
         for state in states_map.keys():
           if state not in valid_states:
             print('Invalid translation state', state)
             return False
-      elif self.units in device_map.keys():
+      elif self.units in translation_map.keys():
```
berkoben, line 188: see above

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -182,9 +200,16 @@ def _ValidateTranslation(self):
             print('Invalid translation unit', unit)
             return False
       else:
-        print('Translation has improper keys:', device_name)
+        print('Translation has improper keys:', field_name)
         return False
 
+    #check if the rest of the fields not included are optional
+    for optional_field_name in all_fields_dict.values():
+      if not optional_field_name.optional:
```
berkoben, line 209: doesn't this just fail any time a type has an optional field?

berkoben (changes requested):
Didn't look at the tests.  Most of this is fine for code we're going to toss in a week, but I think there are a couple logic errors that need fixing.

tools/validators/instance_validator/validate/entity_instance.py
charbull: moved down, if it is udmi compliant not need to check the units and state

tools/validators/instance_validator/validate/entity_instance.py
charbull: no the opposite

tools/validators/instance_validator/validate/entity_instance.py
charbull: can you provide an example please?
Not sure I follow

tools/validators/instance_validator/validate/entity_instance.py
charbull: how do we extract that? with string manipulation?
I prefer to do this in the universe config in a different PR?

tools/validators/instance_validator/validate/entity_instance.py
charbull: if it is not optional it will return

charbull:
PTAL

tools/validators/instance_validator/validate/entity_instance.py
charbull: how do we extract that? with string manipulation?
I prefer to do this in the universe config in a different PR?

PR #117, opened 2020-09-10
adding support for units validation on a field

charbull:
adding validation support for the units validation on a given field


tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -118,6 +130,24 @@ def GetEntityType(self, namespace_name, typename):
...
+		"""
+    if not self.unit_universe_reverse_map:
+      print('UnitUniverse undefined in ConfigUniverse')
+      return None
+    subfields = field_name.split('_')
+    # if the last element is numeric need to remove it
+    while subfields[-1].isnumeric():
+      subfields.pop()
+      
+    if subfields[-1] in ['setpoint', 'sensor', 'accumulator', 'command']:
```
berkoben, line 147: If you filter point types at all I would do it in the reverse (excluding the ones that you know for sure will never have a unit.  That makes it more future proof.

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -118,6 +130,24 @@ def GetEntityType(self, namespace_name, typename):
...
+      measurement_subfield = subfields[-2] # access the measurement_type subfield
+      return self.unit_universe_reverse_map.get(measurement_subfield)
```
berkoben, line 149: (non-blocking) Strictly speaking there's a bug here:  If a subfield is redefined locally this lookup could fail in odd ways.  We really need to work namespacing into the "correct" solution

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -62,15 +62,27 @@ class ConfigUniverse(findings_lib.Findings):
...
   def __init__(self, entity_type_universe, field_universe, subfield_universe,
                state_universe, unit_universe):
-    super(ConfigUniverse, self).__init__()
+    super().__init__()
     self.entity_type_universe = entity_type_universe
     self.field_universe = field_universe
     self.subfield_universe = subfield_universe
     self.state_universe = state_universe
     self.unit_universe = unit_universe
+    self.unit_universe_reverse_map = self._ArrangeUnitsByMeasurement()
```
berkoben, line 71: keep this in the units universe.  pulling it up breaks encapsulation.

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -118,6 +130,24 @@ def GetEntityType(self, namespace_name, typename):
       return None
     return self.entity_type_universe.GetEntityType(namespace_name, typename)
 
+  def GetUnitsMapByMeasurement(self, field_name):
```
berkoben, line 133: this should really be a method on the units universe.  Whether or not you want a proxy to it here is up to you.

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -62,15 +62,27 @@ class ConfigUniverse(findings_lib.Findings):
...
     # temporary placeholder for instance validator
     self.connections_universe = None
 
+  def _ArrangeUnitsByMeasurement(self):
+    if not self.unit_universe:
+      print('UnitUniverse undefined in ConfigUniverse')
+      return None
+    
+    unitsByMeasurement = dict()
+    units = self.unit_universe.GetUnitsMap('')
```
berkoben, line 81: label the bug here.  The ontology technically supports units below the global namespace.  This does not support that behavior correctly.

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -168,45 +168,38 @@ def _ValidateTranslation(self):
...
-
-        unit_values_map = units_map[self.values]
-        for unit in unit_values_map.keys():
-          if unit not in valid_units:
-            print('Invalid translation unit', unit)
-            return False
-      else:
-        print('Translation has improper keys:', field_name)
-        return False
+      if valid_units:
```
berkoben, line 179: note: this needs to be more specific to differentiate between no-units-universe and this field does not have any valid units (i.e. empty vs. none)

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -168,45 +168,38 @@ def _ValidateTranslation(self):
...
+        # the UDMI format
+        elif self.unit_values in translation_map.keys():
+          unit_values_map = translation_map[self.unit_values].data
+          for unit in unit_values_map.keys():
+            if unit not in valid_units:
+              print('Invalid translation unit:', unit)
+              print('Field translation: ', field_name.data)
+              return False
+
+      if valid_states:
```
berkoben, line 196: ditto here for no valid states vs no state universe.

tools/validators/instance_validator/tests/entity_instance_test.py
```diff
@@ -201,6 +201,20 @@ def testValidateMultipleCompliantTranslationWithNamespaceOther(self):
     if not instance_lighting.IsValidEntityInstance():
       self.fail('exception incorrectly raised')
 
+  def testValidateMultipleCompliantTranslationWithIdenticalTypes(self):
+    parsed = instance_parser.parse_yaml(os.path.join(_TESTCASE_PATH,
+                                                       'GOOD',
+                                                       'good_translation'
+                                                       '_identical.yaml'))
```
berkoben, line 208: (non-blocking) When we redo this, we should separate parsing of the different syntaxes and the validation of the contents by creating a native object as a go-between.

tools/validators/instance_validator/tests/fake_instances/GOOD/good_translation_units_and_states.yaml
```diff
@@ -18,9 +18,9 @@ CHWS_WDT-17:
   translation:
     return_water_temperature_sensor:
       present_value: "points.return_water_temperature_sensor.present_value"
-      states:
-        OPEN: "1"
-        CLOSED: "2"
+      unit_values:
+        degrees_celsius: "degC"
+        degrees_fahrenheit: "degF"
```
berkoben, line 23: do these changes change any tests?  (I see a few test files changed but only one test change)

berkoben (changes requested):

tools/validators/instance_validator/tests/fake_instances/GOOD/good_translation_units_and_states.yaml
charbull: not really, the way the tests are written is to just check if an exception is raised or not, need to update those as well

tools/validators/instance_validator/tests/entity_instance_test.py
charbull: agreed

tools/validators/instance_validator/validate/entity_instance.py
charbull: if there is no units universe can someone put units in the config file?

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
charbull: I see, we would need to update in case we added new ones with no units.

charbull: [`status`, `label`, `mode`, `counter`, 'timestamp`]

charbull:
merging and addressing the comments in the next PR due to an urgent need

PR #119, opened 2020-09-11
Add validation on states of fields

charbull:


tools/validators/ontology_validator/yamlformat/validator/field_lib.py
```diff
@@ -113,7 +121,7 @@ def __init__(self,
                parent_namespace=None,
                local_subfields=None,
                local_states=None):
-    super(FieldFolder, self).__init__(folderpath, base_lib.ComponentType.FIELD)
+    super().__init__(folderpath, base_lib.ComponentType.FIELD)
```
berkoben, line 124: what's going on here? doesn't this break inheritance?

tools/validators/ontology_validator/yamlformat/validator/field_lib.py
```diff
@@ -80,7 +80,15 @@ def IsFieldDefined(self, fieldname, namespace_name):
     """
     return fieldname in self._namespace_map.get(namespace_name, set())
 
+  def GetFieldsMap(self, namespace_name):
+    """Returns the fields defined within namespace.
```
berkoben, line 84: be more specific about _exactly_ what is returned, including exactly what the keys are expected to be in the map. (because python)

berkoben (changes requested):
This might work on the current common cases, but it's starting to really make a mess.  I don't think we should make any more changes after this until we roll back and rewrite everything related to unit and state validation.  There's no way we're going to find all the bugs in this approach.

charbull:
Merging for now due to an urgent need we will refactor accordingly

PR #123, opened 2020-09-18
Add check for abstract types

charbull:


tools/validators/instance_validator/tests/entity_instance_test.py
```diff
@@ -82,6 +82,24 @@ def testValidateBadEntityNamespace(self):
     if instance.IsValidEntityInstance():
       self.fail('exception not raised')
 
+
+  def testValidateBadAbstractType(self):
```
berkoben, line 86: test names should be more declarative: testValidateRejectsUseOfAbstractType or similar

tools/validators/instance_validator/tests/fake_instances/BAD/bad_abstract_type.yaml
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ZONE is an abstract type.
+ZONE-105:
+  type: HVAC/ZONE
+  id: CDM/f7ed1725-53f9-4978-98c7-648f5d66e98e
```
berkoben, line 18: probably good to use something fake here...

tools/validators/instance_validator/tests/entity_instance_test.py
```diff
@@ -82,6 +82,24 @@ def testValidateBadEntityNamespace(self):
...
+    parsed = instance_parser.parse_yaml(
```
berkoben, line 87: this test is brittle because it uses a real abstract type that could be deleted.  A better version would define both the instance to check and the type to check against, however since we have to rewrite a bunch of this it's not a blocker right now.

berkoben (approved):
a few suggestions.  Nothing blocking.

tools/validators/instance_validator/tests/entity_instance_test.py
charbull: good catch

tools/validators/instance_validator/tests/entity_instance_test.py
charbull: sgtm

tools/validators/instance_validator/tests/fake_instances/BAD/bad_abstract_type.yaml
charbull: Fakes are already tested, they are not found in the universe

PR #128, opened 2020-10-05
Update to Fields and Subfields for Alarms and Measured Statuses

tasodorff:
This updates many subfields and fields for APAC deployments and contains two major changes:
1. The subfield `alarm` is promoted from a measurement_descriptor to a point_type, thus enablinig measurement_descriptors to be used to be used.
2. For many measured values all that is returned is a boolean (e.g. a flowrate is measured as being greater than a threshold or not), and so additional fields have been added where the measurement descriptor exists for status point types.

Additional fields and subfields are added for specific APAC deployments.


ontology/yaml/resources/states/states.yaml
```diff
@@ -43,3 +44,13 @@ MATCHES: Two or more things match.
 SUBSCRIPTION_FAILED: The calendar events subscription has failed.
 UNKNOWN: The state is unknown.
 WAITING_FOR_RESPONSE: The calendar events synchronization process is awaiting response.
+
+# Added for measured statues (e.g. flowrate_status)
+PRESENT: "Something is currently materialized physically (e.g. flowrate_status is present)."
```
berkoben, line 49: being a little more verbose may add clarity: 
(e.g. in flowrate_status, flow is present)

ontology/yaml/resources/states/states.yaml
```diff
@@ -43,3 +44,13 @@ MATCHES: Two or more things match.
...
+ABSENT: "Something is currently not materialized (e.g. flowrate_status is absent). "
```
berkoben, line 50: ditto

berkoben (approved):
This is good to go EXCEPT you'll need to go through the field name change process for the existing fields you changed.

PR #130, opened 2020-10-07
Pubsub client 

charbull:
reads from pusbusb and prints payload


charbull:
@nkilmer take a look please, will do the test after (ignore doc strings for now)

tools/validators/instance_validator/validate/subscriber.py
```diff
@@ -0,0 +1,57 @@
...
+		audience = "https://pubsub.googleapis.com/google.pubsub.v1.Subscriber"
+		
+		credentials = jwt.Credentials.from_service_account_info(service_account_info, audience=audience)
+		
+		sub_client = pubsub_v1.SubscriberClient(credentials=credentials)
+		return sub_client
+	
+		
+	
+	def Listen(self, sub_client):
```
nkilmer: I think it may be better for the Subscription object to own the subscription client, rather than having to separately manage the client returned by the Config method and pass it to Listen.

tools/validators/instance_validator/validate/subscriber.py
```diff
@@ -0,0 +1,57 @@
...
+	 subscription_name: name of the subscription.
+	 service_account_info: service account information from the GCP project.
+	 """
+	
+	def __init__(self, subscription_name, service_account_info_json_file):
+		super().__init__()
+		self.subscription_name = subscription_name
+		self.service_account_info_json_file = service_account_info_json_file
+	
+	def Config(self):
```
nkilmer: This method seems to mainly be about the creation of a client, so Config may be a misnomer.

tools/validators/instance_validator/validate/subscriber.py
nkilmer: Sorry, I misinterpreted the commit structure (still getting used to github reviews). I see both of my points were resolved in later commits. Resolving.

nkilmer (approved):
LGTM, though I may have further thoughts after starting to work with this code. If so, I'll submit pull requests for your review.

charbull:
@nkilmer: TODO(charbull): I will add tests

PR #133, opened 2020-10-10
Fix multi-entity link bug.

berkoben:
This bumps out the check that all the required fields are covered until after all the lined devices are processed.


berkoben:
charbull tested and this works

PR #135, opened 2020-10-19
Update generate_universe.py

charbull:


charbull:
factoring minor things

PR #151, opened 2020-11-17
delete BUILD file for copybara

kswang2400:


charbull:
deleting the files to import to g3

PR #157, opened 2020-11-24
Partial fix to door inheritance

berkoben:


charbull (changes requested):
type validator error

berkoben:
Don't know what was wrong with this, but I couldn't get it to parse.  replaced it with another PR

PR #165, opened 2020-12-08
Fix file path issues in ontology validator unit tests

nkilmer:


tools/validators/ontology_validator/yamlformat/tests/presubmit_validate_types_lib_test.py
```diff
@@ -86,8 +85,8 @@ def setUp(self):
     self.good2_file = base_lib.PathParts(
         root=self.base_dir, relative_path='GOOD/entity_types/good2.yaml')
 
-    self.good1_depot_path = os.path.join('//depot/google3', RESOURCE_PATH,
-                                         'GOOD/entity_types/good1.yaml')
+    self.good1_depot_path = path.join('//depot/google3', RESOURCE_PATH,
```
charbull, line 88: I think we can get rid of this?

charbull (approved):
nit: we can remove the google3 ones?

PR #168, opened 2020-12-11
Support validation of multiple input files

nkilmer:


tools/validators/instance_validator/tests/handler_test.py
```diff
@@ -45,6 +44,17 @@ def testValidateOneBuildingExistFails(self):
       instance_handler = handler.ValidationHelper(args)
```
nkilmer, line 44: A question about this test (github won't let me add a comment on line 40/39): why is the test file GOOD/good_links.yaml in the GOOD directory if it causes a SyntaxError to be raised?

I noticed that there is an entity commented out in good_links.yaml as well. Is that the reason the error is raised? Can we remove the commented lines and move the file to BAD/bad_links.yaml?

charbull: +1

tools/validators/instance_validator/validate/handler.py
```diff
@@ -164,9 +167,10 @@ def _ParseArgs(self, args):
 
     parser.add_argument(
         '-i', '--input',
-        dest='filename',
+        action='append',
+        dest='filenames',
         required=True,
-        help='Filepath to YAML building configuration',
+        help='Filepaths to YAML building configurations',
```
charbull, line 173: nit: is it just one filepath which contains various config fileS?

nkilmer: No, I enabled the 'append' option for the parameter, so the user can run it like: python instance_validator.py -i file1 -i file2

Even if only one input file is provided, the parameter will still be resolved to a list with one element.

Thanks for reminding me to update the README.

I wonder if we should add another --input-dir parameter that is mutually exclusive with --input which behaves in the way you were thinking.

tools/validators/instance_validator/tests/handler_test.py
nkilmer: Cool, will do this in a separate PR.

tools/validators/instance_validator/validate/handler.py
charbull: I wonder if we need both options.
Providing a directory which picks up all the files might be the simplest?

Just thinking about maintainability and keeping it simple for the user. 
What do you think?

charbull: .

charbull (approved):

tools/validators/instance_validator/validate/handler.py
nkilmer: I think I like having the option for both individual file validation and validation of an entire directory. It should make certain test / development tasks easier if we allow both (e.g. if the user has a particular file they're trying to fix and doesn't want to run the validator on the whole directory as they're iterating on it).

I opened a bug to add the new --input-dir parameter.

PR #171, opened 2020-12-11
Rename test resource good_links.yaml to bad_links_missing_building.yaml

nkilmer:


nkilmer:
Hmm... so it seems that another test failed because it was relying on good_links.yaml. I feel like something's wrong in the test design when one test uses the file as a success case and another uses it as a failure case. I'm also confused because the test in handler_test.py expects (correctly) the parser to raise a SyntaxError on this file, yet the test that's now failing in entity_instance_test.py expects the file to be parsed successfully...

charbull:
@nkilmer tests are failing PTAL

tools/validators/instance_validator/tests/handler_test.py
```diff
@@ -39,8 +39,8 @@ def testValidateOneBuildingExist(self):
 
   def testValidateOneBuildingExistFails(self):
     with self.assertRaises(SyntaxError):
-      # there is missing building type in the test file
-      input_file = path.join(_TESTCASE_PATH, 'GOOD', 'good_links.yaml')
+      input_file = path.join(
+        _TESTCASE_PATH, 'BAD', 'bad_links_missing_building.yaml')
```
charbull, line 43: just rename to missing_building.yaml?

charbull (approved):

charbull:
Yes there is a test which depends on the good yaml
So you can just add a bad_missing_building.yaml file and leave the other one?

charbull:
@nkilmer can you please take a look

nkilmer:
> @nkilmer can you please take a look

Fixed.

charbull (approved):

PR #180, opened 2020-12-16
Pull/179

charbull:


tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -18,14 +18,15 @@
...
 
+from absl.testing import absltest
 from validate import entity_instance
 from validate import instance_parser
 from validate import telemetry_error
 from validate import telemetry_validator
-from absl.testing import absltest
-from os import path
+
+# import threading
```
berkoben, line 29: is this obsolete?  if so, remove.  Otherwise explain

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -114,11 +115,16 @@ def _ParseTranslation(self, translation_body):
...
-    for field_name in translation_body.keys():
-      if isinstance(translation_body[field_name], str):
+    for std_field_name in translation_body.keys():
+      if isinstance(translation_body[std_field_name], str):
         continue
 
-      ft = translation_body[field_name]
+      ft = translation_body[std_field_name]
+
+      raw_field_name = str(ft[PRESENT_VALUE])\
```
berkoben, line 124: I don't follow this at all.  Explain what's happening in a comment plz?

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -43,6 +44,10 @@
                                        'message_attributes_SDC_EXT-17.json')
 _MESSAGE_ATTRIBUTES_3 = json.load(open(_MESSAGE_ATTRIBUTES_PATH_3))
 
+_MESSAGE_ATTRIBUTES_PATH_FACILITIES = \
```
berkoben, line 47: revert all unnecessary explicit linebreaks.

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -114,11 +115,16 @@ def _ParseTranslation(self, translation_body):
 
     translation = {}
 
-    for field_name in translation_body.keys():
-      if isinstance(translation_body[field_name], str):
+    for std_field_name in translation_body.keys():
```
berkoben, line 118: are we requiring fully qualified field names in the config?  If not, how are we normalizing?

tools/validators/instance_validator/validate/field_translation.py
```diff
@@ -18,13 +18,15 @@ class FieldTranslation(object):
...
   Args:
-    field_name: name of the field the translation corresponds to
+    std_field_name: standard name of the field in the ontology
+    raw_field_name: name of the field sent in the telemetry payload
     units: dictionary from standard units to expected telemetry units
     states: dictionary from standard states to expected telemetry states
   """
 
-  def __init__(self, field_name, units, states):
+  def __init__(self, std_field_name, raw_field_name, units, states):
```
berkoben, line 27: can we use type hints or explain the data type in the header comment?  Specifically, I was thinking about field names, which are often tuples in the ontology validator, and it wasn't immediately obvious if a imilar patter was being used here.  (Note: it might make sense to share the same tuple type)

tools/validators/instance_validator/validate/handler.py
```diff
@@ -133,13 +133,14 @@ def TelemetryValidationCallback(self, validator):
...
 
+    print('Generating validation report ...')
     report = ''
 
     if not validator.AllEntitiesValidated():
       report += 'No telemetry message was received for the following entities:'
       report += '\n'
       for entity_name in validator.GetUnvalidatedEntities():
-        report += '  ' + entity_name + '\n'
+        report += '  ' + str(entity_name) + '\n'
```
berkoben, line 143: is this coercion?  If so, I'd prefer using the explicit method on the object.

tools/validators/instance_validator/validate/telemetry_error.py
```diff
@@ -40,7 +40,7 @@ def GetPrintableMessage(self):
     """Returns a human-readable message that explains this error."""
 
     msg = "Error for entity [" + self.entity + "]"
-    if self.point:
-      msg += ", point [" + self.point + "]"
-    msg += ": " + self.message
+    if str(self.point):
```
berkoben, line 43: coercion again?

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -36,12 +39,14 @@ class TelemetryValidator(object):
 
   def __init__(self, entities, timeout, callback):
```
berkoben, line 40: without args defined in the header comment I don't know what entities is.  Please add docs.

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -36,12 +39,14 @@ class TelemetryValidator(object):
...
     super().__init__()
-    self.entities = entities
+    self.entities = dict(filter((lambda e: e[1].translation),
```
berkoben, line 42: this feels like a magic number to me.  #Define?

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -80,44 +86,54 @@ def ValidateMessage(self, message):
...
       return
     self.validated_entities[entity_name] = True
 
     entity = self.entities[entity_name]
-    for point_name, translation in entity.translation.items():
-      if point_name not in tele.points.keys():
+
+    print('Validating telemetry message for entity: {0}'.format(entity_name))
+
+    for translation in entity.translation.values():
```
berkoben, line 104: field_translation for clarity.

berkoben (changes requested):

tools/validators/instance_validator/tests/telemetry_validator_test.py
charbull: done

tools/validators/instance_validator/validate/entity_instance.py
charbull: this part extracts the foo part in the present_value: "points.foo.present_value"

```
CHWS_WDT-17:
  type: HVAC/CHWS_WDT
  id: CDM/123456
  translation:
    return_water_temperature_sensor:
      present_value: "points.return_water_temperature_sensor.present_value"
      ```

File: tools/validators/instance_validator/tests/telemetry_validator_test.py (hunk 3 above)
Comment (charbull) [reply tools/validators/instance_validator/tests/telemetry_validator_test.py:47]: what do you think about bringing all in 3p then `g4 fix` then update to github?

File: tools/validators/instance_validator/validate/entity_instance.py (hunk 4 above)
Comment (charbull) [reply tools/validators/instance_validator/validate/entity_instance.py:118]: We are expecting the following:
```
CHWS_WDT-17:
  type: HVAC/CHWS_WDT
  id: CDM/123456
  translation:
    return_water_temperature_sensor:
      present_value: "points.foo.present_value"
      ```
      
std_field_name == return_water_temperature_sensor
raw_field_name == foo

tools/validators/instance_validator/validate/field_translation.py
charbull: yes, I added a todo in this file, the next PR I am refactoring the code to extract the yaml parsing, and validation to delegate/include in each class. Will do a pass on the types.

tools/validators/instance_validator/validate/handler.py
charbull: Can you detail what you mean?
entity_name is of type YAML

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: done

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: entities: a dict with entity_name as a key and EntityIntance as value.
e[1] is kist accessing the entityInstance obj

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: done

charbull (commented):
PTAL

tools/validators/instance_validator/validate/entity_instance.py
berkoben: that name isn't fully qualified though.  (qualified would be /return_water_temperature_sensor).  We generally don't require full qualification in configurations unless it would be ambiguous, but that means the code needs to normalize everything to not get confused (which is why we use [this tuple](https://github.com/google/digitalbuildings/blob/master/tools/validators/ontology_validator/yamlformat/validator/entity_type_lib.py#L34) when referencing fields in types.

tools/validators/instance_validator/validate/entity_instance.py
berkoben: why is foo important?  The payload could have any arbitrary json path as the present_value and still be valid.

tools/validators/instance_validator/validate/field_translation.py
berkoben: bugid? (or issue ID if preferred)

also arg description should begin with the data type.  Ex: "String. Standard field name in the ontology"

tools/validators/instance_validator/validate/handler.py
berkoben: as far as I can tell it's a string.  https://github.com/google/digitalbuildings/blob/11bd795774d30c2ba10b10d164a7ef280ade2184/tools/validators/instance_validator/validate/handler.py#L47

What I meant by coercion, is that if the thing is a class that has a .name() method, for instance.  I would prefer using the method to get the string rather than depending on the builtin __str__() to get a string value.  Here it appears that there's a different problem.  The GetUnvalidatedEntities() method doesn't actually return entities so its name should really be changed to be something like GetUnvalidatedEntityNames().  Then this would all be clearer.

tools/validators/instance_validator/validate/telemetry_validator.py
berkoben: got it.  The internet did not yield a more elegant way to do it.  

It would be clearer, however if the varname was more explicit about what it was:  ex: entities_with_translations

berkoben (changes requested):

tools/validators/instance_validator/validate/handler.py
```diff
@@ -200,11 +207,11 @@ def _ParseArgs(self, args):
 
     parser.add_argument(
         '-r', '--report-filename',
-        dest = 'report_filename',
-        required = False,
-        default = None,
-        help = 'Filename for the validation report',
-        metavar = 'report-filename')
+        dest='report_filename',
```
nkilmer, line 210: nit/question: Is param_name=param_value the correct style for Python named parameters? It looks wrong to me without whitespace but I can adjust if that's the convention.

tools/validators/instance_validator/validate/point.py
```diff
@@ -25,4 +25,12 @@ class Point(object):
   def __init__(self, point_name, value):
     super().__init__()
     self.point_name = point_name
-    self.present_value = value
+    self.present_value = self._ToString(value)
+
+  def _ToString(self, value):
```
nkilmer, line 30: I think people (like me) might be confused about why this isn't using the built-in __str__ method. Perhaps calling it _ValueToString will clear up that confusion.

tools/validators/instance_validator/validate/point.py
```diff
@@ -25,4 +25,12 @@ class Point(object):
...
+    if isinstance(value, bool):
+      return '{}'.format(value).lower()
```
nkilmer, line 32: In this case, str(value) might be better, since there's no additional string content - you're just converting the value to a string.

tools/validators/instance_validator/validate/point.py
```diff
@@ -25,4 +25,12 @@ class Point(object):
...
+
+    if isinstance(value, int) or isinstance(value, float):
+      return '{}'.format(value)
```
nkilmer, line 35: See above.

tools/validators/instance_validator/validate/telemetry.py
```diff
@@ -77,8 +77,12 @@ def _parse_data(self, data):
     version = json_object[VERSION]
     timestamp = json_object[TIMESTAMP]
     points = {}
+    if POINTS not in json_object.keys():
```
nkilmer, line 80: As mentioned above, I like the idea of making this as generic as possible by not assuming that 'points' is the top-level key for the JSON telemetry values.

Instead we can validate this kind of structural convention at the entity_instance level and then rely on the YAML translation to tell us where to find the points in the JSON payload.

tools/validators/instance_validator/validate/telemetry_error.py
nkilmer: Should probably be:

if self.point is not None

... if we want to be explicit and not rely on Python's interpretation of value truthiness. (Though I think that is considered Pythonic.)

tools/validators/instance_validator/validate/telemetry_error.py
```diff
@@ -39,8 +39,8 @@ def __eq__(self, other):
   def GetPrintableMessage(self):
     """Returns a human-readable message that explains this error."""
 
-    msg = "Error for entity [" + self.entity + "]"
-    if self.point:
-      msg += ", point [" + self.point + "]"
-    msg += ": " + self.message
+    msg = "- entity [" + self.entity + "]"
```
nkilmer, line 42: I'm in favor of removing the "Error for" prefix to decrease redundancy in the error report. However, can we also ditch the "- " prefix and rely on the report generation code to format properly? This message might be used in other contexts where we don't want to have any prefix.

tools/validators/instance_validator/validate/telemetry_error.py
```diff
@@ -39,8 +39,8 @@ def __eq__(self, other):
...
+    if str(self.point):
+      msg += ", point [" + str(self.point) + "]"
```
nkilmer, line 44: self.point should always be a string or None according to the docstring, so str() isn't necessary here.

I like string.format for this kind of thing though. (Sorry, I was still de-rusting my Python when I wrote this code.)

tools/validators/instance_validator/validate/telemetry_error.py
nkilmer, line 42: nit: Maybe change the double quotes to single quotes to match our quote usage in other files? (My fault, I hadn't yet noticed the quote convention when I wrote this code.)

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -22,12 +22,20 @@
...
 from validate import telemetry_error
+from validate import telemetry_warning
 
 DEVICE_ID = 'deviceId'
 
 class TelemetryValidator(object):
   """Validates telemetry messages against a building config file.
 
+  Attributes;
+    entities: a dict with entity_name as a key and EntityIntance as value.
```
nkilmer, line 33: entity_name doesn't mean anything to the reader here.

You could say the key is the entity name as a string.

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -22,12 +22,20 @@
...
+    timeout: the max time the validator must read messages from pubsub.
+    callback: the method called by the pubsub listener upon receiving a msg.
```
nkilmer, line 35: Actually, this is the callback that the validator calls when it's done (either because messages for all entities were seen or because the timeout duration was reached).

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -22,12 +22,20 @@
...
+    validated_entities: list of entities validated during the timeout duration.
```
nkilmer, line 36: These two (validated_entities, validation_errors) should be private. I think that's indicated by prefixing them with an underscore and not mentioning them in the attributes doc?

tools/validators/instance_validator/validate/telemetry_validator.py
nkilmer: The technique explained here seems more readable:

https://stackoverflow.com/questions/23862406/filter-items-in-a-python-dictionary-where-keys-contain-a-specific-string

tools/validators/instance_validator/validate/telemetry_warning.py
```diff
@@ -0,0 +1,46 @@
...
+            self.message == other.message)
+
+  def GetPrintableMessage(self):
+    """Returns a human-readable message that explains this warning."""
+
+    msg = "- entity [" + self.entity + "]"
+    if str(self.point):
+      msg += ", point [" + str(self.point) + "]"
+    msg += ": " + self.message+"\n"
+    return msg
```
nkilmer, line 46: nit: missing EOF newline

nkilmer (changes requested):

tools/validators/instance_validator/validate/handler.py
nkilmer: @berkoben Oops, my review took longer than an hour.

Sorry if we overlapped on some comments.

tools/validators/instance_validator/validate/telemetry_validator.py
nkilmer: nit: typo for EntityIn_s_tance

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -105,9 +112,13 @@ def _CreateEntityInstances(yaml_filename):
 
 # A set of two test entities with identical points.
 _ENTITIES_3_4 = _CreateEntityInstances('good_translation_identical.yaml')
+
```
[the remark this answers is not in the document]
charbull: done

tools/validators/instance_validator/tests/telemetry_validator_test.py
charbull: updated the explicit linebreaks

tools/validators/instance_validator/validate/entity_instance.py
charbull: b/176094783 created will get to it in the next PR while refactoring entity instance to make it more oop

tools/validators/instance_validator/validate/field_translation.py
charbull: TODO + b/176095085

tools/validators/instance_validator/validate/handler.py
charbull: updated the name and replaced with ,format

tools/validators/instance_validator/validate/handler.py
charbull: pylint was complaining so I changed it similar to the ones above

tools/validators/instance_validator/validate/telemetry_error.py
charbull: this blows up without the Str as it a YAML string this is how I got around it.

tools/validators/instance_validator/validate/telemetry_error.py
charbull: it doesn't work without the str, the error message is trying to cast Yaml to str

tools/validators/instance_validator/validate/entity_instance.py
charbull: Discussed offline with Keith, created the following b/176097512

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: updated to entities_with_translations, I prefer the lambda one :)

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -105,9 +112,13 @@ def _CreateEntityInstances(yaml_filename):
...
 _ENTITY_NAME_3 = 'SDC_EXT-17'
 _ENTITY_NAME_4 = 'SDC_EXT-18'
 
+_ENTITIES_5 = _CreateEntityInstances('good_translation_multi_states.yaml')
```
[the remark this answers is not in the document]
charbull: done

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -248,11 +253,20 @@ def ValidationCallback(validator):
...
       self.assertTrue(validator.AllEntitiesValidated())
 
-    validator = telemetry_validator.TelemetryValidator(
-      _ENTITIES_1, 1, ValidationCallback)
+    validator = telemetry_validator\
+      .TelemetryValidator(_ENTITIES_1, 1, ValidationCallback)
 
     validator.ValidateMessage(_MESSAGE_GOOD)
 
+  def testTelemetryValidatorOnMultiStateValidated(self):
```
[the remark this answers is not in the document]
charbull: updated

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -248,11 +253,20 @@ def ValidationCallback(validator):
...
+    def ValidationCallback(validator):
+      self.assertEmpty(validator.GetErrors())
+      self.assertTrue(validator.AllEntitiesValidated())
+
+    validator = telemetry_validator \
+      .TelemetryValidator(_ENTITIES_5, 1, ValidationCallback)
+
+    validator.ValidateMessage(_MESSAGE_MULTIPLE_STATES)
```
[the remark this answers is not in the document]
charbull: done

tools/validators/instance_validator/validate/handler.py
```diff
@@ -22,11 +22,11 @@
...
 from validate import instance_parser
 from validate import subscriber
-
-# Default timeout duration for telemetry validation test
 from validate import telemetry_validator
+from datetime import datetime
 
-DEFAULT_TIMEOUT = 600
+# Default timeout duration for telemetry validation test
+DEFAULT_TIMEOUT = 300
```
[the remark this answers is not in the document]
charbull: done

tools/validators/instance_validator/validate/point.py
charbull: done

tools/validators/instance_validator/validate/point.py
charbull: done

tools/validators/instance_validator/validate/telemetry.py
charbull: created a todo for the rest,
this check is diferent, often we are receiving from live data error messages with strings (logs) instead of points. 
this check is necessary here

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: I kept entity name for now as we defined it here for the user: https://github.com/google/digitalbuildings/blob/master/ontology/docs/building_config.md

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: added _ and removed from attributes

tools/validators/instance_validator/validate/telemetry_warning.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Container for a telemetry validation error."""
+
+class TelemetryWarning(object):
```
berkoben, line 17: do we want to link into the findings model here?

berkoben (changes requested):

tools/validators/instance_validator/validate/entity_instance.py
charbull: updated in the new push

tools/validators/instance_validator/validate/field_translation.py
charbull: updated the bug with the latest comment and added a todo

tools/validators/instance_validator/tests/fake_instances/GOOD/good_translation_multi_states.yaml
```diff
@@ -0,0 +1,39 @@
...
+      present_value: points.run_status.present_value
+      states:
+        OFF: 0
+        ON: 1
+    speed_percentage_command:
+      present_value: points.supply_fan_speed_percentage_command.present_value
+      units:
+        key: pointset.points.supply_fan_speed_percentage_command.units
+        values:
+          percent: '%'
```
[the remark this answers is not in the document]
charbull: done

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: updated

tools/validators/instance_validator/validate/telemetry_warning.py
charbull: done

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -372,13 +373,13 @@ def _ValidateLinkStatesMatch(self, universe, source_field, target_field):
...
     if source_states != target_states:
-      print('State mismatch in link from {0} to {1}'\
+      print('State mismatch in link from {0} to {1}'
             .format(source_field, target_field))
       return False
     return True
 
 
-  def IsValidEntityInstance(self, universe = None, entity_instances = None):
+  def IsValidEntityInstance(self, universe=None, entity_instances=None):
```
berkoben, line 382: removing the spaces, I'm pretty sure, is incorrect.

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -107,6 +115,9 @@ def _CreateEntityInstances(yaml_filename):
 _ENTITIES_3_4 = _CreateEntityInstances('good_translation_identical.yaml')
 _ENTITY_NAME_3 = 'SDC_EXT-17'
 _ENTITY_NAME_4 = 'SDC_EXT-18'
+# A test entity to make sure that boolean are retrieved and validated.
```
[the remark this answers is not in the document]
charbull: done

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -44,26 +42,27 @@ class TelemetryValidator(object):
 
   def __init__(self, entities, timeout, callback):
     super().__init__()
-    self.entities = dict(filter((lambda e: e[1].translation),
+    self.entities = dict(filter((lambda entities_with_translations:
```
berkoben, line 45: this is backward.  The output is entities with translations.  The filter is taking entities and removing those without translations.

tools/validators/instance_validator/validate/telemetry_warning.py
charbull: yes want to refactor and merge TelemetryWarning and TelemetryError into TelemetryReportElement? and we can have two or more levels or messages: warning, errors (for now)

berkoben (changes requested):

tools/validators/instance_validator/validate/entity_instance.py
charbull: I have this error from the linter:
```
validate/entity_instance.py:382:43: C0326: No space allowed around keyword argument assignment
  def IsValidEntityInstance(self, universe = None, entity_instances = None):
                                           ^ (bad-whitespace)
validate/entity_instance.py:382:68: C0326: No space allowed around keyword argument assignment
  def IsValidEntityInstance(self, universe = None, entity_instances = None):
                                                                    ^ (bad-whitespace)
```

tools/validators/instance_validator/validate/handler.py
charbull: I have this error from the linter:

validate/entity_instance.py:382:43: C0326: No space allowed around keyword argument assignment
  def IsValidEntityInstance(self, universe = None, entity_instances = None):
                                           ^ (bad-whitespace)
validate/entity_instance.py:382:68: C0326: No space allowed around keyword argument assignment
  def IsValidEntityInstance(self, universe = None, entity_instances = None):
                                                                    ^ (bad-whitespace)

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -37,13 +37,14 @@ class TelemetryValidator(object):
...
-    callback: callback function to be called when the validation finishes
+    callback: callback function to be called either because messages for all
+      entities were seen or because the timeout duration was reached
   """
 
   def __init__(self, entities, timeout, callback):
     super().__init__()
-    self.entities = dict(filter((lambda entities_with_translations:
-                                 entities_with_translations[1].translation),
+    self.entities = dict(filter((lambda entities:
```
berkoben, line 46: self.entities should be self.entities_with translations, no?

tools/validators/instance_validator/validate/telemetry_error.py
charbull: no longer needed the str part as the var is parsed earlier

tools/validators/instance_validator/validate/telemetry_validator.py
charbull: oh I see
not my preferred to use super long variable names, updated

berkoben (approved):

PR #199, opened 2021-01-08
merge the schema of the translation in building config schema

charbull:


tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -33,7 +33,19 @@
 """Schema separately parses translation to account for multiple valid formats
 github.com/google/digitalbuildings/blob/master/ontology/docs/building_config.md
 #defining-translations"""
-_TRANSLATION_SCHEMA = syaml.Str() | syaml.Any()
+_TRANSLATION_SCHEMA = syaml.Str() | syaml.MapPattern(
+    syaml.Str(),
```
berkoben, line 37: doesn't the external pipe invalidate the stricter validation?

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -33,7 +33,19 @@
...
+    syaml.Str() | syaml.Map({'present_value': syaml.Str(),
```
berkoben, line 38: same here.

tools/validators/instance_validator/validate/instance_parser.py
berkoben, line 37: Can we be stricter about the keys? We have rules for case sensitivity and format for fields.

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -33,7 +33,19 @@
...
+                             syaml.Optional('states'): syaml.MapPattern(
+                                 syaml.Str(), syaml.Str()),
```
berkoben, line 40: this is always CAP[_LTRS]

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -42,34 +54,19 @@
                            syaml.Map({
                                'type': syaml.Str(),
```
berkoben, line 55: ditto on strictness.

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -42,34 +54,19 @@
...
                                'id': syaml.Str(),
-                               syaml.Optional('connections'):
-                                 syaml.MapPattern(syaml.Str(),
-                                                  syaml.Str()) |
-                                 syaml.Seq(
-                                     syaml.MapPattern(syaml.Str(),
-                                                      syaml.Str())),
+                               syaml.Optional('connections'): syaml.MapPattern(
+                                   syaml.Str(), syaml.Str()) | syaml.Seq(
+                                       syaml.MapPattern(syaml.Str(),
```
berkoben, line 59: and here.  And Why the | ?  I don't know what the other syntax option is.

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -42,34 +54,19 @@
...
+                                                        syaml.Str())),
                                syaml.Optional('links'): syaml.MapPattern(
                                    syaml.Str(),
                                    syaml.MapPattern(syaml.Str(), syaml.Str())),
```
berkoben, line 63: and here. these must both be valid fields.

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -84,22 +81,19 @@ def _ValidateEntityWithSchema(content, schema):
...
 
-    # TODO(b/176502336): merge the schemas to have one validation
     # check the translation
     top_name = parsed.keys()[0]
     if _TRANSLATION in parsed.data[top_name].keys():
       translation = parsed[top_name][_TRANSLATION]
-      translation.revalidate(_TRANSLATION_SCHEMA)
 
       # if translation is not UDMI compliant
+      # We need to test this seperatly as strictyaml does not support a
```
berkoben, line 90: separately

berkoben (changes requested):

tools/validators/instance_validator/validate/instance_parser.py
charbull: the Str is for the key word `Compliant` which is the case with `translation: compliant`
Strictyaml does not allow you to do this:

`Compliant` | | syaml.MapPattern( ...

See output error in the description below on why we need to validate Comlpiant seperatly

tools/validators/instance_validator/validate/instance_parser.py
charbull: done

tools/validators/instance_validator/validate/instance_parser.py
charbull: added a todo b/166472270 to revisit with the connections. Ray was trying things out, dont remember why we have a Seq here. We need to clean it up with the connections work

tools/validators/instance_validator/validate/instance_parser.py
charbull: the first element str() is for the case when `translation: compliant`.
Strictyaml dont let you do `string_value` | Map thats why it is generic str() here then we do a validation at the end

tools/validators/instance_validator/validate/instance_parser.py
charbull: are you referring to states?

we can use an enum: https://hitchdev.com/strictyaml/using/alpha/scalar/enum/
but lets keep it lower case for simplicity

tools/validators/instance_validator/validate/instance_parser.py
charbull: are you referring to connections?

tools/validators/instance_validator/validate/instance_parser.py
berkoben: Enums are always CAPS, including in our docs.

tools/validators/instance_validator/validate/instance_parser.py
berkoben: in the link, both the values must be valid fields

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -42,34 +54,20 @@
                            syaml.Map({
                                'type': syaml.Str(),
                                'id': syaml.Str(),
-                               syaml.Optional('connections'):
-                                 syaml.MapPattern(syaml.Str(),
-                                                  syaml.Str()) |
-                                 syaml.Seq(
-                                     syaml.MapPattern(syaml.Str(),
-                                                      syaml.Str())),
+                               # TODO(b/166472270): revisit connections
```
berkoben, line 57: Always explain the "why". ex: "revisit connection syntax validation.  Current code might not follow the spec."

berkoben (changes requested):

tools/validators/instance_validator/validate/instance_parser.py
charbull: done

tools/validators/instance_validator/validate/instance_parser.py
charbull: done

tools/validators/instance_validator/validate/instance_parser.py
berkoben: this key can also use the field regex, no?

tools/validators/instance_validator/validate/instance_parser.py
```diff
@@ -27,13 +27,26 @@
 _IGNORE_PATTERN = re.compile(r'^(\W)*#|\n')
 # number of entities to validate per batch
 _ENTITIES_PER_BATCH = 1
-_COMPLIANT = 'COMPLIANT'
+_COMPLIANT_REGEX = u'^COMPLIANT$'
 _TRANSLATION = 'translation'
+_FIELD_REGEX = u'(^[a-z]+[a-z0-9]*(?:_[a-z]+[a-z0-9]*)*)(?:_[0-9]+)*$'
```
berkoben: remove first capture group as well.  not needed in this application

berkoben (changes requested):

berkoben (approved):

PR #216, opened 2021-03-05
AU-SYD-ODI types

IBMSService:
First types including Delamtic lighting types, heat thermal meter, three AHU types and supporting abstract and telemetry field changes.


IBMSService:
@googlebot I signed it!

IBMSService:
@googlebot I signed it!

PR #263, opened 2021-05-12
automation tools from xls to udmi and dbo

pisuke:
Hi @charbull @tasodorff and @berkoben.
Here is a PR to include the draft automation tools to the tools folder.
Very much work in progress as the template spreadsheet is still in flux, but let me know your comments about location, code standards, etc. Thanks!


tools/automation/building_blocks/device.py
```diff
@@ -0,0 +1,90 @@
...
+  information needs to be stored.
+
+  Call to_dictionary() to retrieve all the data.
+"""
+
+import abc
+
+class Device:
+  
+  class Connection:
```
kswang2400, line 14: Connection, Translation, Link classes seem like they could just be tuples? if these classes are meant to be extended, maybe define one class and inherit to be a little more DRY?

tools/automation/building_blocks/device.py
```diff
@@ -0,0 +1,90 @@
...
+  def __init__(self, device_name, device_type, device_id):
+    self._name = device_name
+    self._type = device_type
+    self._id = device_id
+    self._connections = {}
+    self._translation = {}
+    self._links = {}
+  
+  def populate_connections(self, name, value):
+    self._connections.update(self.Connection(name, value).to_dictionary())
```
kswang2400, line 50: if a connection is just one key value pair, is there a benefit to creating the Connection object first? something like `self._connections[name] = value` seems cleaner and easier to read

same with `populate_translations` and `populate_links`

tools/automation/building_blocks/device.py
```diff
@@ -0,0 +1,90 @@
...
+
+  """
+    Call this to retrieve a dictionary representing this data.
+
+    If connections are empty, don't add the connection value to dictionary.
+
+    If translations are empty, don't add the translation value to dictionary.
+
+    If links are empty, don't add the link value to dictionary.
+  """
```
kswang2400, line 66: nit: docstring inside function

https://www.python.org/dev/peps/pep-0257/#id17

tools/automation/building_blocks/device.py
```diff
@@ -0,0 +1,90 @@
...
+  def to_dictionary(self):
+    return_dictionary = {
+      self._name: {
+        "type": self._type,
+        "id": self._id,
+      }
+    }
+
+    if len(self._connections) > 0:
+      return_dictionary[self._name].update({
```
kswang2400, line 76: any reason to prefer calling update rather than just object assignment? seems a bit unnecessary to create a new temporary dictionary especially if it's only one key

```
return_dictionary[self._name]["connections"] = self._connections
```

also, is there a reason we don't just set connections automatically to self._connections without checking for length? 

```
    return_dictionary = {
      self._name: {
        "type": self._type,
        "id": self._id,
        "connections": self._connections,
        "translations": self._translations,
        "links": self._links,
      }
    }
```

tools/automation/building_blocks/floor.py
```diff
@@ -0,0 +1,57 @@
...
+
+class Floor:
+  
+  def __init__(self, floor_name, floor_id):
+    self._name = floor_name
+    self._id = "FACILITIES/{0}".format(floor_id)
+    self._connections = {}
+  
+  def populate_connections(self, building_name):
+    self._connections.update({building_name: "CONTAINS"})
```
kswang2400, line 20: nit: for single key updates maybe prefer assignment

`self._connections[building_name] = "CONTAINS"`

more readable and slightly better performance

https://stackoverflow.com/a/15456218

tools/automation/sections/building_section.py
```diff
@@ -0,0 +1,33 @@
...
+  
+  Holds logic for reading input files to retrieve the data for a building.
+"""
+
+from building_blocks.building import Building
+
+
+class DBOBuildingSection:
+
+  def __init__(self, files):
```
kswang2400, line 14: nit: can we annotate what `files` are?

https://www.python.org/dev/peps/pep-3107/

tools/automation/sections/device_section.py
```diff
@@ -0,0 +1,212 @@
...
+  They also hold the logic for how to read the input files in order to
+  retrieve the correct data for a device.
+"""
+
+import re
+import abc
+
+from building_blocks.device import Device
+
+"""
```
kswang2400, line 15: nit: docstring inside class declaration

tools/automation/sections/device_section.py
```diff
@@ -0,0 +1,212 @@
...
+class DeviceSection:
+
+  def __init__(self, files):
+    self._devices = []
+
+  # Logic for reading the excel data to fill up device data.
+  @abc.abstractmethod
+  def _create_device(self, data):
+    pass
```
kswang2400, line 26: nit: maybe raise a NotImplementedError instead of passing silently

https://docs.python.org/3/library/exceptions.html#NotImplementedError

tools/automation/sections/device_section.py
```diff
@@ -0,0 +1,212 @@
...
+
+    return state_dict
+
+  def _fill_device_connections(self, row, device):
+    feeds = row[self._site_model_columns.CONNECTION_FEEDS]
+    controls = row[self._site_model_columns.CONNECTION_CONTROLS]
+
+    if not feeds.isspace() and len(feeds) > 0:
+      feeds = self._get_devices_from_string(feeds)
+      for i in feeds:
```
kswang2400, line 80: nit: "i" seems to imply index here but I think it's supposed to be a connection name, maybe prefer "name" instead

tools/automation/xls2dbo.py
```diff
@@ -0,0 +1,142 @@
...
+"""
+class YAML:
+
+  # Default location for all YAML files.
+  OUTPUT_PATH = "./dbo.yaml"
+
+  # Write the YAML structure and contents into the dbo file.
+  @abc.abstractmethod
+  def create_yaml_file(self):
+    None
```
kswang2400, line 46: nit: raise NotImplementedError

tools/automation/xls2dbo.py
```diff
@@ -0,0 +1,142 @@
...
+  f1 = Figlet(font='standard')
+  print(f1.renderText('xls2DBO'))
+
+def main():
+      
+  show_title()
+
+  parser = argparse.ArgumentParser()
+  group = parser.add_mutually_exclusive_group()
+  group.add_argument("-v", "--verbose", action="store_true", default=False, help="increase the verbosity level")
```
kswang2400, line 122: nit: we should probably run a linter on this before landing

kswang2400 (changes requested):
mostly stylistic comments but PR is large so probably worth discussing before approval :)

charbull:
@pisuke @tasodorff should we keep this in the background or you want to bring loadboy in?

PR #276, opened 2021-06-30
Issue 260 units

pisuke:
I've put together a pull request to include the BACnet units into the DBO list of units.
The reason for this is getting alignment between units in UDMI and DBO.
Let me know if you have any comments. 
@grafnu should this be successful, should we also change the `units.json` file in the UDMI schema to have the same snakecase way of writing units?
As a side note, there is a very complete units ontology: [QUDT](http://www.qudt.org/2.1/catalog/qudt-catalog.html). It would be good to do an alignment between DBO and QUDT.
Also, DBO units miss to define conversion factors between standard unit and alternative ones: is this something that will be added in the future?
Thanks for having a look!

### DIFF
--- a/ontology/yaml/resources/units/units.yaml
+++ b/ontology/yaml/resources/units/units.yaml
@@ -61,6 +61,7 @@ power:
 - kilovolt_amperes_reactive
 pressure:
 - pascals: STANDARD
+- kilopascals
 - inches_of_water
 - pounds_force_per_square_inch
 resistance:
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -385,6 +385,30 @@ DFVSC:
   implements:
   - OPERATIONAL
 
+DFVSC2X:
+  description: "Variable speed control for discharge fans for 2 separate zones."
+  is_abstract: true
+  opt_uses:
+  - discharge_fan_current_sensor_1
+  - discharge_fan_power_sensor_1
+  - discharge_fan_speed_frequency_sensor_1
+  - discharge_fan_speed_percentage_sensor_1
+  - discharge_fan_lost_power_alarm_1
+  - discharge_fan_current_sensor_2
+  - discharge_fan_power_sensor_2
+  - discharge_fan_speed_frequency_sensor_2
+  - discharge_fan_speed_percentage_sensor_2
+  - discharge_fan_lost_power_alarm_2
+  uses:
+  - discharge_fan_speed_percentage_command_1
+  - discharge_fan_run_status_1
+  - discharge_fan_run_command_1
+  - discharge_fan_speed_percentage_command_2
+  - discharge_fan_run_status_2
+  - discharge_fan_run_command_2
+  implements:
+  - OPERATIONAL
+
 
 DFVSFC:
   id: "3819485279274663936"
@@ -1319,6 +1343,24 @@ CHWZTC:
   implements:
   - CONTROL
 
+CHWZTC2X:
+  description: "Chilled water valve control on zone side (ZTC) for two separate zones. Chilled water valve controls to the worst zone."
+  is_abstract: true
+  opt_uses:
+  - discharge_air_temperature_sensor_1
+  - discharge_air_temperature_sensor_2
+  - leaving_cooling_coil_temperature_sensor
+  - cooling_thermal_power_capacity
+  - chilled_water_flowrate_sensor
+  - chilled_water_valve_percentage_sensor
+  uses:
+  - zone_air_temperature_setpoint_1
+  - zone_air_temperature_setpoint_2
+  - zone_air_temperature_sensor_1
+  - zone_air_temperature_sensor_2
+  - chilled_water_valve_percentage_command
+  implements:
+  - CONTROL
 
 # DX Control
 
@@ -2090,6 +2132,26 @@ HT2ZC:
   implements:
   - CONTROL
 
+HT2XZTC2X:
+  description: "Two separate heating sections going to two different zones on the device, with independent heating control."
+  is_abstract: true
+  opt_uses:
+  - heating_percentage_command_1
+  - discharge_air_temperature_sensor_1
+  - heating_percentage_command_2
+  - discharge_air_temperature_sensor_2
+  uses:
+  - zone_air_temperature_setpoint_1
+  - zone_air_temperature_sensor_1
+  - heater_run_command_1
+  - heater_run_status_1
+  - zone_air_temperature_setpoint_2
+  - zone_air_temperature_sensor_2
+  - heater_run_command_2
+  - heater_run_status_2
+  implements:
+  - CONTROL
+
 HT3ZC:
   id: "3504555462265667584"
   description: "Two gas or electric heater control on zone side (HSP, DSP)."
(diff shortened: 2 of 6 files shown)


ontology/yaml/resources/units/units.yaml
```diff
@@ -12,58 +12,319 @@
...
+-	volts_per_meter: STANDARD
+
+electricinductance:
+-	henrys: STANDARD
+
+electricphasedifference:
+-	degrees_phase: STANDARD
+
+electromagneticfieldstrength:
+-	amperes_per_meter: STANDARD
```
grafnu, line 72: spacing seems inconsistent -- why is this line (72) different than say the next one (75)?

grafnu (commented):
RE: UDMI, once this is in, I likely think units.json will just go away and be replaced by a simple regex check on a valid unit structure (snake case).  No need/desire to explicitly list them there as well.

charbull (changes requested):
We are relying on QUDT already.
We are going through an internal clean up to update the units internally first then export them to github.
Ideally revert the deletions/renaming and keep only the additions as they will be easier to update internally.

Thank you

ontology/yaml/resources/units/units.yaml
pisuke: Thanks, it was an issue with tabs being kept after a copy/paste from the BACnet units specs. Corrected now, thanks for spotting it!

pisuke:
> RE: UDMI, once this is in, I likely think units.json will just go away and be replaced by a simple regex check on a valid unit structure (snake case). No need/desire to explicitly list them there as well.

makes perfect sense, thanks for clarifying

pisuke:
> We are relying on QUDT already.
> We are going through an internal clean up to update the units internally first then export them to github.
> Ideally revert the deletions/renaming and keep only the additions as they will be easier to update internally.
> 
> Thank you

@charbull I've addressed this comment and reverted the deletions and renaming, so there should be just additions now.
 I don't agree with some of the quantity names (level vs length, I've decided to keep both), specificenthalpy could just be enthalpy and some others, but we could discuss this once you have done your internal units update.

Any reason why the quantities don't keep underscore characters to separate words? They are a bit difficult to read.

charbull:
@pisuke can you fix the validation errors here please: https://github.com/google/digitalbuildings/pull/276/checks?check_run_id=3021112121

charbull (changes requested):
@pisuke  please fix the errors in the validation : https://github.com/google/digitalbuildings/pull/276/checks?check_run_id=3021112121

pisuke:
> @pisuke please fix the errors in the validation : https://github.com/google/digitalbuildings/pull/276/checks?check_run_id=3021112121

@charbull OK, I've added the quantities descriptions as requested by the validation process and the validation checks now pass.
However, I disagree with forcing singularity of unit, and for this reason I've commented the "duplicates" I would like to suggest to introduce. 
For instance, meters at the moment a unit for "level" but I think it should also be recorded as a unit for "length".
"no_unit" is a special case of adimensionality to measure powerfactor, but it will be handy for any adimensional quantity.
Thoughts?

charbull:
@tsodorff can you please take a look at this?

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -173,22 +173,63 @@ descriptor:
...
+  electriccapacitance: "The ratio of the change in an electric charge in a system to the corresponding change in its electric potential."
+  electricconductance: "The degree to which an object conducts electricity, calculated as the ratio of the current which flows to the potential difference present"
+  electricconductivity: "The degree to which a specified material conducts electricity, calculated as the ratio of the current density in the material to the electric field which causes the flow of current."
+  electriccurrentdensity: "The amount of electric current flowing per unit cross-sectional area of a material."
+  electricfield: "A region around a charged particle or object within which a force would be exerted on other charged particles or objects."
+  electricinductance: "The property of an electric conductor or circuit that causes an electromotive force to be generated by a change in the current flowing."
+  electricphasedifference: "The time interval by which a wave leads by or lags by another wave."
+  electromagneticfieldstrength: "Quantitative expression of the intensity of an electric field at a particular location"
```
tasodorff, line 190: is this an electric field or a magnetic field? they measure different things; electric field is in V/m and magnetic field is in A/m.

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -173,22 +173,63 @@ descriptor:
...
+  electromagneticmoment: "Measure of the magnetic strength of a magnet or current-carrying coil, expressed as the torque produced when the magnet or coil is set with its axis perpendicular to unit magnetic flux density."
```
tasodorff, line 191: magnetic, not electromagnetic

tasodorff (changes requested):
Some comments on differentiating between magnetic and electric in electromagnetic keywords. Please be clear about if things are electric or magnetic in nature (e.g. which type of field it is).

ontology/yaml/resources/units/units.yaml
```diff
@@ -12,58 +12,315 @@
...
+- ampere_square_meters: STANDARD
+
+electricresistivity:
+- ohm_meters: STANDARD
+
 energy:
 - joules: STANDARD
 - kilowatt_hours
+- btus
+- kilobtus
```
biotvandenberg, line 84: Should the kilo* and mega* units be consistent with underscores?  e.g. this has kilobtus and mega_btus.
Should it be one of:
- kilo_btus and mega_btus or
- kilobtus and megabtus

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -173,22 +173,63 @@ descriptor:
...
   current: "Flow of electric charge."
+  density: "The quantity of mass per unit volume of a substance."
+  electriccapacitance: "The ratio of the change in an electric charge in a system to the corresponding change in its electric potential."
+  electricconductance: "The degree to which an object conducts electricity, calculated as the ratio of the current which flows to the potential difference present"
+  electricconductivity: "The degree to which a specified material conducts electricity, calculated as the ratio of the current density in the material to the electric field which causes the flow of current."
+  electriccurrentdensity: "The amount of electric current flowing per unit cross-sectional area of a material."
+  electricfield: "A region around a charged particle or object within which a force would be exerted on other charged particles or objects."
+  electricinductance: "The property of an electric conductor or circuit that causes an electromotive force to be generated by a change in the current flowing."
+  electricphasedifference: "The time interval by which a wave leads by or lags by another wave."
```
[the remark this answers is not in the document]
pisuke: thanks for the suggestion, description updated to "angular difference"

ontology/yaml/resources/subfields/subfields.yaml
pisuke: updated to magnetic

ontology/yaml/resources/subfields/subfields.yaml
pisuke: updated to electric

ontology/yaml/resources/units/units.yaml
pisuke: thanks for spotting this, for consistency all metric prefixes have been replaced without the underscore

pisuke:
I've addressed all the suggestions, it should be ready to be merged.
For info, another good source of units and quantities information is this https://github.com/ryantenney/gnu-units/blob/master/units.dat
It also includes conversion factors.

pisuke:
@charbull I think I've addressed all the requests, can this PR now be merged?

charbull (changes requested):
@pisuke can you update to resolve the conflicts please?

mschulze17:
#414 has been merged. @charbull, @pisuke - can this PR be addressed now?

PR #279, opened 2021-06-30
Update setup.py

charbull:
ruamel dependency


charbull:
forcing this one to fix ruamel.yaml dependency

PR #280, opened 2021-06-30
Update setup.py

charbull:
removing ruamel dependency and putting strictyaml to 1.1.1


charbull:
forcing it, the error will be handled in the next pr

PR #295, opened 2021-07-28
Add initial interface to tools

trav3711:
Opening a PR for my initial interface wrapper for the validator universe.

### DIFF
--- a/tools/rdf_generator/setup.py
+++ b/tools/rdf_generator/setup.py
@@ -25,6 +25,6 @@ setup(
     author_email='charbelk@google.com',
     description='',
     packages=find_packages(),
-    install_requires=['absl-py', 'pyglib', 'ruamel.yaml', 'rdflib'],
+    install_requires=['absl-py', 'pyglib', 'ruamel.yaml', 'rdflib==5.0.0'],
     python_requires='>=3.6',
 )


trav3711:
@googlebot I signed it!

charbull:
@trav3711 please sync with base branch, I had to set a version to fix the integration tests.
@tasodorff fyi

trav3711:
@charbull Which branch is the base branch? Are you saying I should try to merge with master again?

tools/interface/tests/ontology_interface_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+class OntologyTest(absltest.TestCase):
+
+    def setUp(self):
+        super(OntologyTest, self).setUp()
+        #self.basedir = RESOURCE_PATH
+        yaml_files = external_file_lib.RecursiveDirWalk(validator_constants.ONTOLOGY_ROOT)
+        self.ontology = ontology_interface.Ontology(yaml_files)
+
+    def testGetFieldsForTypeName(self):
```
berkoben, line 24: It is more readable, IMO if the tests are smaller and test one thing at a time.  If this test failed, it would be very difficult to determine what condition was unmet.

berkoben (changes requested):
Commented mainly on the ontology_interface file.  I'd suggest you slim this PR down to ONLY include the Ontology class and tests for that class

tools/interface/ontology_interface.py
```diff
@@ -0,0 +1,89 @@
...
+    def get_fields_for_type_name(self, namespace: str, typename: str, required_only: bool = False) -> list():
+        """gets a list of fields for a given typename within a namespace
+
+           Args:
+                namespace: a string indication the namespace of the entity type
+                typename: a string indicating the name of entity type
+                required_only: when true will return only required fields for a given type
+
+           Returns:
+                result_fields: a list of fieldwrapper NamedTuples where the first tuple element is
```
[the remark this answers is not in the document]
trav3711: OptWrapper felt a little too bulky as all I really need are the field name and its optionality

tools/interface/ontology_interface.py
```diff
@@ -0,0 +1,89 @@
...
+                result_fields: a list of fieldwrapper NamedTuples where the first tuple element is 
+                the field name and the second element is a boolean indicating whether the field is required
+        """
+        pass
+     
+    def get_entities_from_fields(self, namespace: str, real_field_list: list, general_type: str = None, match_type: str = None) -> list():
+        """gets a list of Entitytype objects matching a given a namespace and a list of field names
+
+           Args:
+                namespace: a string indicating the namespace name of the fields
```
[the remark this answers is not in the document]
trav3711: The goal is for the user to be able to pass in a list of field names as strings, and to convert that list to a list of FieldParts or OptWrappers a namespace is required. Thus a namespace needs to be passed into the function. Do you have any ideas to get around this @berkoben

tools/interface/ontology_interface.py
```diff
@@ -0,0 +1,89 @@
...
+                match_field_list: a list of field names
+                general_type: a string indicating a general type name to narrow return results
```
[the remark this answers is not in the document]
trav3711: A namespace is included in the method arguments.

tools/interface/ontology_interface.py
```diff
@@ -0,0 +1,89 @@
...
+
+            Returns:
+                entities: a list of EntityType objects matching the provided list of field names
+        """
+        pass
+
+    def validate_type(self, typename: str) -> bool:
```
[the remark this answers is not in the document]
trav3711: Change to using an EntityType object instead of a string

tools/interface/ontology.py
```diff
@@ -0,0 +1,61 @@
+import sys
+from typing import List, NamedTuple
+
+sys.path.insert(1, '../validators/ontology_validator/')
+
+from yamlformat.validator.entity_type_lib import EntityType
+from yamlformat.validator.presubmit_validate_types_lib import ConfigUniverse
+from yamlformat import constants
+
+StandardField = NamedTuple('StandardField', [('namespace', str), ('name', str), ('is_required', bool)])
```
berkoben, line 10: "StandardField" is a misnomer for this thing. A "standard field" can't have optionality because it's defined outside of the context of an entitytype.  It's also ambiguous whether or not an increment can be part of the name.  

I think the approach that might be easier to use and more flexible in the future is to create an `EntityTypeField` class that has properties and accessor methods for all of the relevant concepts:
-namespace
-name
-increment
-optionality

In the API you can follow the convention that a specified property is a strict "must" have and an unspecified property (==None) is a wildcard.   For example:

`EntityType(namespace = 'HVAC', name='run_status')`  would match`run_status` as an optional or required field.  It would also match `run_status_1`.  

`EntityType(namespace = 'HVAC', name='run_status', increment='', is_optional=False)` would only match required fields having `run_status` as the name.

You can similarly have a `StandardField` object with only `namespace`, `name` if you need it.  You can then use inheritance to do nice stuff like any `EntityTypeField` is also a `StandardField`.  For example, the `EntityTypeField` representing HVAC/run_status_1, optional is also the Standard Field HVAC/run_status.

tools/interface/ontology.py
```diff
@@ -0,0 +1,61 @@
...
+                general_type: a string indicating a general type name to narrow return results
+
+            Returns:
+                entities: a list of EntityType objects matching the provided list of fields
+        """
+        pass
+
+    def IsTypeValid(self, entity_type: EntityType) -> bool:
+        """Validates the existence of a type in the ontology"""
+        pass
```
berkoben, line 51: If you already have an EntityType object, can't you just call IsValid() on it?  Maybe I'm missing something.

berkoben (changes requested):
I commented only on the ontology file.  Please trim this PR down to only the ontology file and files containing any new classes that it references.  If the PR gets too big my review turnarounds are going to get real slow.  Also the more we try to cover at once, the more rework we'll end up having to do.

tools/interface/ontology.py
```diff
@@ -0,0 +1,61 @@
...
+           Args:
+                entity_type: An EntityType object to get fields for 
+                required_only: when true will return only required fields for a given type
+
+           Returns:
+                result_fields: a list of StandardField tuples
+        """
+        pass
+     
+    def GetEntitiesFromFields(self, field_list: List[StandardField], general_type: str = None) -> List[EntityType]:
```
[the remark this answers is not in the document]
trav3711: Since this is where  matching will occur, optionality won't play a role in this method. The default behavior will be to ignore optional fields.

tools/interface/ontology.py
trav3711: I like the idea of having an `EntityTypeField` class extended from some base field class. What stopped me from doing this before was that the `Field` class already exists and creating another similar class would be redundant and could lead to performance issues.

tools/interface/ontology.py
```diff
@@ -7,34 +7,37 @@
...
-            self.universe = A ConfigUniverse object detailing the various universes in the ontology
+            universe = A ConfigUniverse object detailing the various universes in the ontology
+
+       Returns:
+            An instance of Ontology class
     """
 
     def __init__(self, universe: ConfigUniverse):
         super().__init__()
         self.universe = universe
```
berkoben, line 26: typo

tools/interface/ontology.py
```diff
@@ -46,16 +49,50 @@ def GetEntitiesFromFields(self, field_list: List[StandardField], general_type: s
...
+    Attributes:
+        namespace: the name of the namespace as a string
+        name: the field name as a string
+
+    returns:
+        An instance of the StandardField class
 
+    """
+
+    def __init__(self, namespace_name: str, field_name: str):
```
berkoben, line 74: sticking with the convention in https://docs.google.com/spreadsheets/d/1ePTg7HVv-eUU1Lc60dW3tpmaPwJMUy0ZvJRmNaOsMME/edit#gid=0  I would call this "standard_field__name"

tools/interface/ontology.py
```diff
@@ -46,16 +49,50 @@ def GetEntitiesFromFields(self, field_list: List[StandardField], general_type: s
...
+        super().__init__()
+        self.namespace = namespace_name
+        self.name = field_name
```
berkoben, line 77: This will feel like way overkill now, but I promise the pattern will be valuable later:  Make the properties _private and implement getter methods.  This allows you to mess with the value that the user actually gets at a later time without having to change any dependent code.   (use the "correct" name in the getter as well, ex: getStandardFieldName() - this way you'll get nice inheritance behavior)

tools/interface/ontology.py
```diff
@@ -46,16 +49,50 @@ def GetEntitiesFromFields(self, field_list: List[StandardField], general_type: s
...
+
+    Attributes:
+        increment: the increment of the field under a type
+        is_required: optionality of the field relative to it's Entity Type# Configure the Azure provider
+
+    Returns:
+        An instance of the EntityTypeField class
+    """
+
+    def __init__(self, increment: str = '', is_required: bool):
```
berkoben, line 95: the init needs arguments for the namespace and standard field name, then you want to call super with these arguments.

tools/interface/ontology.py
```diff
@@ -7,34 +7,37 @@
...
+                entity_type_name: name of the entity type as a string
                 required_only: when true will return only required fields for a given type
 
            Returns:
                 result_fields: a list of StandardField tuples
         """
         pass
      
-    def GetEntitiesFromFields(self, field_list: List[StandardField], general_type: str = None) -> List[EntityType]:
+    def GetEntityTypesFromFields(self, field_list: List[StandardField], general_type: str = None) -> List[EntityType]:
```
berkoben, line 40: This can be an exercise for later, but it probably makes sense to have your own local version of an entitytype.

berkoben (changes requested):
almost there.

tools/interface/api.py
```diff
@@ -0,0 +1,24 @@
+import sys
```
berkoben, line 1: you can remove this file from the PR to keep it tidy (since this doesn't do anything useful right now)

tools/interface/ontology.py
```diff
@@ -0,0 +1,116 @@
+import sys
```
berkoben, line 1: make the file path for this tools/explorer/lib/

tools/interface/tests/ontology_interface_test.py
```diff
@@ -0,0 +1,127 @@
+import sys
```
berkoben, line 1: this file should be tools/explorer/tests/ontology_test.py

berkoben (changes requested):
on tests:  Can you slim down the test ontology to a smaller number of elements?  

high level:  Can you make the parent for this code tools/explorer with two subs lib/ and tests/

PR #296, opened 2021-07-28
Update setup.py

charbull:
fix rdflib version


charbull:
fix python rdflib version

PR #308, opened 2021-08-12

charbull:
PiperOrigin-RevId: 390472619


charbull:
bringing ids in from type generator

PR #314, opened 2021-08-18
Add testing to Ontology explorer

trav3711:


berkoben (changes requested):
Please add the code that's being tested so the automated testing doesn't fail.

PR #322, opened 2021-09-02
Add ontology.py testing

trav3711:
Opening a PR to add ontology.py logic and testing with the necessay dependencies.

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/AHU.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/AHU.yaml
@@ -3019,6 +3019,7 @@ AHU_BSPC_BYPSSPC_DX2SC_ECON_EFSS_HTSC_SFSS:
   - SFSS
 
 AHU_BSPC_DX4SC_ECON_EFSS_EFVSC_SARC_SFSS_SFVSC_SSPC_SSPCSCM_STCSCM:
+  id: "4539990850909765632"
   description: "Multi-zone AHU with four-stage DX cooling and with supervisor control mode types (machine learning)."
   is_canonical: true
   implements:
@@ -3027,6 +3028,7 @@ AHU_BSPC_DX4SC_ECON_EFSS_EFVSC_SARC_SFSS_SFVSC_SSPC_SSPCSCM_STCSCM:
   - STCSCM
 
 AHU_BSPC_DX2SC_ECON_EFSS_SFSS_SFVSC_SSPC_SSPCSCM_STCSCM:
+  id: "9692108824621613056"
   description: "Multi-zone AHU with two-stage DX cooling and with supervisor control mode types (machine learning)."
   is_canonical: true
   implements:
@@ -3035,6 +3037,7 @@ AHU_BSPC_DX2SC_ECON_EFSS_SFSS_SFVSC_SSPC_SSPCSCM_STCSCM:
   - STCSCM
 
 AHU_BSPC_DXSC_ECON_EFSS_EFVSC_SFSS_SFVSC_SSPC_SARC_SSPCSCM_STCSCM:
+  id: "7170093033294135296"
   description: "Multi-zone AHU with single-stage DX cooling and exhaust/supply fan spped control, and with supervisor control mode types (machine learning)."
   is_canonical: true
   implements:
@@ -3043,6 +3046,7 @@ AHU_BSPC_DXSC_ECON_EFSS_EFVSC_SFSS_SFVSC_SSPC_SARC_SSPCSCM_STCSCM:
   - STCSCM
 
 AHU_BSPC_DXSC_ECON_SFSS_SFVSC_SSPC_SSPCSCM_STCSCM:
+  id: "15294586761070510080"
   description: "Multi-zone AHU with single-stage DX cooling and supply fan spped control, and with supervisor control mode types (machine learning)."
   is_canonical: true
   implements:


tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+    )
+    qualified_fields = entity_type.GetAllFields()
+    field_optwrappers = [
+        entity_type.GetField(field)
+        for field in qualified_fields
+    ]
+    entity_type_fields = [
+        ETF(namespace,
+            field.field.field[0:],
+            not field.optional,
```
berkoben, line 58: generally speaking it's less confusing to keep the same sense for booleans throughout an application.  Is there any reason that we could use True == Optional here too?  (the logic behind making Optional == True is that we wanted things to default to being stricter - if you don't specify, it's required)

tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+  ) -> str:
+    """Determines the closeness of the match between
+    the field list and the entity type
 
     Args:
-       namespace: the name of the namespace as a string.
-       entity_type_name: the name of the entity type as a string.
-       required_only: when true will return only required fields for a given
-         type.
+      field_list: a list of StandardField objects
```
berkoben, line 80: explain the relationship between these two args in the comment

tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+    return entity_type_fields
+
+  #TODO: Expand NONE matching logic
+  def _CreateMatch(
+      self,
+      field_list: List[StandardField],
+      entity_type: EntityType
+  ) -> str:
+    """Determines the closeness of the match between
+    the field list and the entity type
```
berkoben, line 77: Explain the algorithm for determining closeness and howe that maps to outputs.

tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+
+    entity_field_optwrappers = entity_type.GetAllFields().values()
+    entity_field_set = {
+        StandardField(entity_type.namespace.namespace, ow.field.field)
+        for ow in entity_field_optwrappers
+    }
+    input_field_set = set(field_list)
+
+    if entity_field_set.issubset(input_field_set):
+      if input_field_set.issubset(entity_field_set):
```
berkoben, line 96: Use:
 if not entity_field_set.symmetric_difference(input_field_set):

tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+        match_type = 'EXACT'
```
berkoben, line 97: Given that you're using sets of standard fields, it's possible for a non-exact match to show as exact in the case where one or both entities have incremented fields (run_status_1, run_status_2).  Is this as intended?

tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+            entities: a list of EntityType objects
+            matching the provided list of fields
     """
-    pass
+    match_list = []
+
+    field_list_namespace = {
+        field.GetNamespaceName() for field in field_list
+    }
+    assert len(field_list_namespace) == 1, 'Fields not under one namespace'
```
berkoben, line 130: why is this a problem?

tools/explorer/lib/ontology.py
```diff
@@ -31,35 +32,137 @@ def GetFieldsForTypeName(self,
...
+    namespace = field_list_namespace.pop()
```
berkoben, line 131: this will fail under normal circumstances because nearly all fields are defined in the global namespace (''), but very few types are defined in that namespace.  

I think this matching needs some rethinking.  https://github.com/google/digitalbuildings/blob/617eb8ab4739a0c8dad89ca81d1c210dbd221192/tools/validators/ontology_validator/yamlformat/validator/entity_type_manager.py#L86  implements a completely general greatest-common-subset matching algorithm on the ontology.  Maybe you could adapt this approach for your use case?

berkoben (changes requested):
didn't look at tests yet

tools/explorer/lib/model.py
```diff
@@ -1,8 +1,11 @@
-"""Helper Field model classes for Ontology explorer."""
+"""Helper Field model classes for Ontology explorer"""
+from typing import List
 
+from yamlformat.validator.entity_type_lib import EntityType
 
 class StandardField(object):
-  """A class to represent a generic field without increment or optionality.
+  """
+    A class to represent a generic field without increment or optionality
```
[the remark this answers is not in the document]
trav3711: The Linter didn't throw that error but I'll make the change.

tools/explorer/lib/model.py
```diff
@@ -49,23 +69,90 @@ def __init__(self,
...
+
+  def GetEntityType(self) -> EntityType:
+    """Returns the entity type for a match"""
+    return self._entity_type
+
+  def GetMatchType(self) -> str:
+    """Returns the Match Type for a match"""
+    return self._match_type
+
+def ConvertETFtoSF(field_list: List[EntityTypeField]) -> List[StandardField]:
```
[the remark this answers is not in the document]
trav3711: I'm actually thinkink of scrapping this function because a match object needs to know a fields and optionality and increment to proply apply set operations.

tools/explorer/lib/ontology.py
trav3711: Yeah this would make more sense. I'll make the change.

tools/explorer/lib/ontology.py
trav3711: Changing these two EntityTypeFields will fix this.

tools/explorer/lib/ontology.py
trav3711: All of the fields need to be under the same namespace in order to make set operations. See StandardField.__eq__()

tools/explorer/lib/ontology.py
trav3711: Is this true? I was under the impression that the namespace validator assigned fields to namespaces during field expansion.

PR #343, opened 2021-09-27
Fix regex performance issues.

dplute:
A few of the regex patterns for validating names have unnecessary quantifiers that are causing them to take substantially longer to run.

This patch removes the troublesome quantifiers.  The patterns are logically the same, but they no longer experience excessive backtracking.


charbull (changes requested):
can you please update your branch

charbull (approved):
(I will merge it later, we are synching with the internal code)

PR #348, opened 2021-09-27
Update model classes

trav3711:
I added a bunch of changes to more rigorously define StandardField and EntityTypeField. I also added a Match model class to assist with storing match data in Ontology.GetEntityTypesFromFields().

### DIFF
--- a/tools/explorer/lib/model.py
+++ b/tools/explorer/lib/model.py
@@ -1,12 +1,20 @@
-"""Helper Field model classes for Ontology explorer."""
+"""Helper Field model classes for Ontology explorer"""
+from typing import List
+import re
 
+from yamlformat.validator.entity_type_lib import EntityType
+from yamlformat.validator.field_lib import FIELD_CHARACTER_REGEX
+
+FIELD_INCREMENT_REGEX = re.compile(r'((?:_[0-9]+)+)$')
 
 class StandardField(object):
-  """A class to represent a generic field without increment or optionality.
+  """
+    A class to represent a generic field without increment or optionality.
 
     Args:
-        namespace_name: the name of the field's namespace as a string
-        standard_field_name: the name of the field as a string
+        namespace_name: a field's defined namespace as a string
+        standard_field_name: the un-incremented name of the field as a string.
+                             must be lower-case and properly formatted
 
     Attributes:
         namespace: the name of the namespace as a string
@@ -19,26 +27,48 @@ class StandardField(object):
 
   def __init__(self, namespace_name: str, standard_field_name: str):
     super().__init__()
-    self._namespace = namespace_name
+    if not FIELD_CHARACTER_REGEX.match(standard_field_name):
+      raise ValueError('field name is incorrectly formatted')
+    else:
+      self._namespace = namespace_name
     self._name = standard_field_name
 
+  def __hash__(self):
+    return hash((self._namespace, self._name))
+
+  def __eq__(self, other):
+    if not isinstance(other, self.class):
+      raise TypeError('Both objects are not StandardField objects')
+    else:
+      namespace_eq = self._namespace == other.GetNamespaceName()
+      name_eq = self._name == other.GetStandardFieldName()
+      return name_eq and namespace_eq
+
+  def __str__(self):
+    return self._namespace + '/' + self._name
+
   def GetNamespaceName(self) -> str:
-    """Returns namespace variable as a string."""
+    """Returns namespace variable as a string"""
     return self._namespace
 
   def GetStandardFieldName(self) -> str:
-    """Returns the unqualified field name without any increment as a string."""
+    """Returns the unqualified field name
+    without any increment as a string"""
     return self._name
 
-
 class EntityTypeField(StandardField):
-  """A class to represent a field assigned to a type and extends StandardField.
+  """
+  A class to represent a field assigned to a type and extends StandardField
+
+  Args:
+      namespace_name: a field's defined namespace as a string
+      standard_field_name: the name of the field as a string
+      increment: the increment of the field as a string
+      is_optional: optionality of the field as a boolean
 
   Attributes:
       increment: the increment of the field under a type
-      is_required: optionality of the field relative to it's Entity Type#
-                   Configure the Azure provider
-
+      is_optional: optionality of the field relative to it's Entity Type
 
   Returns:
       An instance of the EntityTypeField class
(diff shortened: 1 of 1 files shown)


tools/explorer/lib/model.py
```diff
@@ -47,25 +77,93 @@ class EntityTypeField(StandardField):
...
 
   def GetIncrement(self) -> str:
-    """Returns the EntityType Field's increment as a string."""
+    """Returns the EntityType Field's increment as a string"""
     return self._increment
 
-  def IsRequired(self) -> bool:
-    """Returns the optionality of the field as a boolean."""
-    return self._is_required
+  def IsOptional(self) -> bool:
```
charbull, line 113: unit tests on public methods?

tools/explorer/lib/model.py
```diff
@@ -1,12 +1,20 @@
-"""Helper Field model classes for Ontology explorer."""
+"""Helper Field model classes for Ontology explorer"""
+from typing import List
+import re
 
+from yamlformat.validator.entity_type_lib import EntityType
+from yamlformat.validator.field_lib import FIELD_CHARACTER_REGEX
+
+FIELD_INCREMENT_REGEX = re.compile(r'((?:_[0-9]+)+)$')
```
charbull, line 8: move this to `field_lib`?

tools/explorer/lib/model.py
```diff
@@ -1,12 +1,20 @@
...
 class StandardField(object):
-  """A class to represent a generic field without increment or optionality.
+  """
+    A class to represent a generic field without increment or optionality.
 
     Args:
-        namespace_name: the name of the field's namespace as a string
-        standard_field_name: the name of the field as a string
+        namespace_name: a field's defined namespace as a string
+        standard_field_name: the un-incremented name of the field as a string.
```
charbull, line 16: you can take any standard field name and remove the increment in the class?

tools/explorer/lib/model.py
```diff
@@ -47,25 +77,93 @@ class EntityTypeField(StandardField):
...
+
+class Match(object):
+  """
+  A class to hold the information about a match between a list of
+  EntityTypeFields objects and an Entity Type
+
+  Args:
+    field_list: a list of EntityTypeField objects
+    entity_type: an entity type which implements a subset of field_list
+    match_type: the closeness of a match between field_list and entity type
```
charbull, line 125: is this a string or a num?

charbull (changes requested):
can you add unit tests please

tools/explorer/lib/model.py
trav3711: @charbull what do you mean by this?

tools/explorer/lib/model.py
```diff
@@ -47,25 +77,93 @@ class EntityTypeField(StandardField):
   def __init__(self,
                namespace_name: str,
                standard_field_name: str,
-               is_required: bool,
+               is_optional: bool,
```
kswang2400, line 80: docstring deleted? also semantically seems like this flag is a breaking change?

kswang2400 (approved):
lgtm, but please get approval from @charbull as well

tools/explorer/lib/model.py
trav3711: @kswang2400 This is actually fix from what was a breaking change. Since the FieldParts tuples store optionality where True is optional. I adjusted my class to better fit that tuple. The goal is to be able to convert optwrapper and fieldparts tuples to objects.

charbull:
@trav3711 can you please update the branch.
@dplute can you review please.

tools/explorer/lib/model.py
charbull: your class can accept a standard field name with an increment and clean it up by removing the increment. if you want to be more flexible. but it can also be left as it is.

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
+"""Testing module for model.py"""
```
charbull, line 1: Thank you ! :)

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+from yamlformat.validator.entity_type_lib import EntityType
+from yamlformat.validator.external_file_lib import RecursiveDirWalk
+from yamlformat.validator import presubmit_validate_types_lib
+from yamlformat.validator import namespace_validator as nv
+
+class StandardFieldTest(absltest.TestCase):
+  def setUp(self):
+    self.test_standard_field = StandardField(
+        namespace_name='',
```
charbull, line 18: `foo` better than leaving it blank

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
...
+        standard_field_name='supply_air_flowrate_sensor'
+    )
+
+  def testGetNamespaceName(self):
+    expected_output = ''
```
charbull, line 23: `foo`

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
...
+  def testGetStandardFieldName(self):
+    expected_output = 'supply_air_flowrate_sensor'
+    function_output = self.test_standard_field.GetStandardFieldName()
+
+    self.assertEqual(function_output, expected_output)
+
+class EntityTypeFieldTest(absltest.TestCase):
+  def setUp(self):
+    self.test_entity_type_field = EntityTypeField(
+        namespace_name='',
```
charbull, line 37: anything but empty one :)

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
...
+        standard_field_name='supply_air_flowrate_sensor',
+        is_optional=False,
+        increment=''
```
charbull, line 40: what happens when there is an increment here?

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+  def testStandardizeField(self):
+    expected_output = StandardField(
+        namespace_name='',
+        standard_field_name='supply_air_flowrate_sensor'
+    )
+
+    function_output = StandardizeField(self.test_entity_type_field)
+
+    self.assertEqual(function_output, expected_output)
```
charbull, line 60: thank you for arranging the tests :)

tools/explorer/lib/tests/model_test.py
```diff
@@ -0,0 +1,114 @@
...
+    ]
+    self.test_match = Match(
+        field_list=test_field_list,
+        entity_type=test_entity_type,
+        match_type='EXACT'
+    )
+
+  def testGetFieldList(self):
+    #print(test_constants.ONTOLOGY_ROOT)
+    #print(self.universe.entity_type_universe.type_namespaces_map)
```
charbull, line 88: leftovers

tools/explorer/lib/tests/test_constants.py
```diff
@@ -0,0 +1,10 @@
+"""Constants for the ontology explorer application."""
+
+from os import path
+
+# TODO(b/183433133):move to one constants file after refactoring.
+JOINS = path.join('..', '..', '..', '..')
```
charbull, line 6: you can reuse this one instead of defining a new one?
https://github.com/google/digitalbuildings/blob/master/tools/validators/instance_validator/validate/constants.py

charbull: ah maybe not? it is in a different folder

charbull (changes requested):
minor changes

tools/explorer/lib/tests/model_test.py
trav3711: @charbull '' is the global namespace.

tools/explorer/lib/tests/model_test.py
trav3711: See above

tools/explorer/lib/tests/test_constants.py
trav3711: Yeah I think not. I've tried using the ontology validator constants before and it's easier to just create my own constants since the explorer is a separate package.

tools/explorer/lib/tests/model_test.py
trav3711: Good idea.

tools/explorer/lib/model.py
```diff
@@ -19,26 +25,48 @@ class StandardField(object):
 
   def __init__(self, namespace_name: str, standard_field_name: str):
     super().__init__()
-    self._namespace = namespace_name
+    if not FIELD_CHARACTER_REGEX.match(standard_field_name):
+      raise ValueError('field name is incorrectly formatted')
```
charbull, line 29: please add the namespace and field name in the message returned to the user.

tools/explorer/lib/model.py
```diff
@@ -47,25 +79,99 @@ class EntityTypeField(StandardField):
...
+        self._namespace,
+        self._name,
+        self._increment,
+        self._is_optional
+    ))
+
+  def __eq__(self, other):
+    if not isinstance(other, self.__class__):
+      raise TypeError(
+          '{str(other)} and {str(self)} must be EntityTypeField objects'
```
charbull, line 104: `EntityTypeField`is a concept that does not exist in our documentation. What does the user need to do when they have this error?

tools/explorer/lib/model.py
charbull: let the author resolve conversations in general please :)

tools/explorer/lib/model.py
```diff
@@ -47,25 +79,99 @@ class EntityTypeField(StandardField):
...
+class Match(object):
+  """
+  A class to hold the information about a match between a list of
+  EntityTypeFields objects and an EntityType object.
+
+  Args:
+    field_list: a list of EntityTypeField objects
+    entity_type: an entity type which implements a subset of field_list
+    match_type: the closeness of a match between field_list and entity type
+    as a string. Current values are: EXACT, CLOSE, INCOMPLETE, NONE
```
charbull, line 133: thank you for adding those !

charbull (approved):
please look at minor comments related to the user messages.

tools/explorer/lib/model.py
trav3711: @Charbull I'm thinking that a human user won't have to deal with these errors. This is a class created purely for data storage, and if these errors are thrown, then it would be through a mistake made in another module.

tools/explorer/lib/model.py
trav3711: no prob

Issue #176, opened 2020-12-15
Instance validator thread keep listening after the report is published


charbull:
**Describe the bug**
Instance validator thread keep listening after the report is published

**To Reproduce**
run the validator on the s2 file

**Expected behavior**
After the report is published the instance validator needs to exit

mschulze17:
@charbull - can we get an update on this Issue?

charbull:
@mschulze17 this has not been fixed, it will be good to bring it to our next sprint.

PR #361, opened 2021-10-04
Fix telemetry validator not exiting after timeout

dplute:
This patch addresses issue #176, where the instance validator does not exit from telemetry validation after the desired timeout.  sys.exit does not work because it only closes the program when it is called from the main thread.  Instead, _thread.interrupt_main raises a KeyboardInterrupt exception on the main thread, which Subscriber.Listen will catch and then gracefully exit.

I don't have any valid pubsub account credentials, so I have only tested this by replacing the Subscriber.Listen call with an infinite loop that sleeps for short periods and catches KeyboardInterrupt exceptions, which allowed the program to exit when the timeout was reached.  This should be tested with a real subscriber client running, but I believe the threading behavior will be the same.

### DIFF
--- a/tools/validators/instance_validator/validate/handler.py
+++ b/tools/validators/instance_validator/validate/handler.py
@@ -17,6 +17,7 @@ from __future__ import print_function
 
 from datetime import datetime
 import sys
+import _thread
 from typing import Callable, Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


charbull (approved):

charbull:
can you please update the branch?

PR #364, opened 2021-10-05
New MAU type

cstirdivant:
@tasodorff  please advise, this is a single-zone unit with air flowrate control (is it ok classify as outside air control? Since outside air and discharge air are one-in-the-same here)


tasodorff (approved):
OAFC in this instance makes sense, since this is just pass-through.

PR #372, opened 2021-10-13
Support writing instance validator results to file

ghost:
This update changes the behavior of the existing --report_filename argument so that the report file will contain all output messages from both instance validation and telemetry validation, instead of just telemetry.

I have moved the telemetry callback to the module scope and removed BuildTelemetryValidationCallback because there is no longer any need for the callback to be a closure.


dplute:
Does anyone have suggestions on dealing with the consider-using-with lint error?  A 'with' block is not really appropriate here since we are not always opening a file, and we're letting python manage it as a replacement stdout.

trav3711:
You can run the linter locally with pylint --rcfile=DigitalBuildints/styles/pylintrc file_name.py

PR #384, opened 2021-10-20
Add argparse to ontology validator

trav3711:
Removing abseil and replacing with argparse for command line arguments for ontology validator.


charbull:
Did it run locally?
Any tests we can add to  avoid a revert ?
Thx

trav3711:
> Did it run locally? Any tests we can add to avoid a revert ? Thx

It ran locally. Let me add a test, but I'm not sure how to test for command line arguments.

PR #388, opened 2021-10-21
New SENSOR type for IoT Pilot

cstirdivant:
New types for NCD vibration and DP sensors. @tasodorff @mschulze17

Vendor documentation for reference:
https://store.ncd.io/product/industrial-wireless-vibration-and-temperature-sensor/
https://ncd.io/wireless-bidirectional-differential-pressure-sensor-product-manual/


tasodorff:
Note: since these apply to HVAC devices they are fine in that namespace.

Issue #399, opened 2021-10-27
Can't send multistate value as string?


a-o-u:
**Describe the bug**
Our gateways are configured to send the multistate value as a string in the telemetry.

When running the telemetry instance validator, sending (multistate) value as a string produces an error.

**To Reproduce**
Instance Config (note that the translation for each of the states is itself):
```
DMP_EDM-17:
  type: HVAC/DMP_EDM
  id: CDM/123456
  cloud_device_id: "foobar"
  translation:
    exhaust_air_damper_command:
      present_value: "points.exhaust_air_damper_command.present_value"
      states:
        OPEN: "OPEN"
        CLOSED: "CLOSED"
    exhaust_air_damper_status: MISSING

UK-LON-S2:
  type: FACILITIES/BUILDING
  id: FACILITIES/123456
```

Telemetry:
```
{
  "timestamp": "2020-10-15T17:21:59.000Z",
  "version": 1,
  "points": {
    "exhaust_air_damper_command": {
      "present_value": "OPEN"
    },
  }
}
```
Error:
```
- entity [DMP_EDM-17], point [points.exhaust_air_damper_command.present_value]: Missing state in telemetry message: None
```
***
The below is also invalid
```
DMP_EDM-17:
  type: HVAC/DMP_EDM
  id: CDM/123456
  cloud_device_id: "foobar"
  translation:
    exhaust_air_damper_command:
      present_value: "points.exhaust_air_damper_command.present_value"
      states:
        OPEN: "1"
        CLOSED: "0"
    exhaust_air_damper_status: MISSING

UK-LON-S2:
  type: FACILITIES/BUILDING
  id: FACILITIES/123456
```

Telemetry:
```
{
  "timestamp": "2020-10-15T17:21:59.000Z",
  "version": 1,
  "points": {
    "exhaust_air_damper_command": {
      "present_value": "1"
    },
  }
}
```
Error:
```
- entity [DMP_EDM-17], point [points.exhaust_air_damper_command.present_value]: Missing state in telemetry message: None
```

**Expected behavior**
No errors

**OS (please complete the following information):**
Windows

**Additional context**
Though this might be intended behaviour? However the building config above does pass the (non-telemetry) instance validator.

Upon inspection it seems like the [_ValueToString method in the point class](https://github.com/google/digitalbuildings/blob/master/tools/validators/instance_validator/validate/point.py#L30) doesn't handle strings?

PR #405, opened 2021-10-27
Allow string multistate telemetry

a-o-u:
For Issue #399 

Also updated tests to check if is multistate mapped in building config


charbull (approved):

charbull:
@a-o-u Thank you !

PR #411, opened 2021-11-02
add entity id to missing GUID warning

trav3711:
Opening PR for adding the entity id to warning message for missing GUID.


charbull (approved):
thank you :)

PR #414, opened 2021-11-10
Redesign unit_lib to support measurement aliases.

dplute:
This branch adds the proposed measurement aliasing feature to the ontology validator.

* Added syntax for measurements that copy another measurement.
* Adjusted UnitUniverse and UnitNamespace methods for better encapsulation.
* Removed the Unit.measurement_type field. Instead, UnitNamespace now maintains a map from measurements to the list of units that belong to it.
* At the end of processing the unit file, each aliased measurement copies the units that were defined in the base measurement.
* 'no_units' is still the only unit name that is allowed to appear more than once in the units YAML file, any other duplication is achieved through measurement aliasing.

### DIFF
--- a/tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
+++ b/tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
@@ -241,6 +241,28 @@ class UnitLibTest(absltest.TestCase):
     self.assertIsInstance(folder.local_namespace.GetFindings()[0],
                           findings_lib.MeasurementAliasIsAliasedError)
 
+  def testUnitFolderAddFromConfigDuplicateAlias(self):
+    doc_1 = {
+      'distance': [
+        {'yards': 'STANDARD'},
+        'inches'
+      ],
+      'level': 'distance',
+    }
+    doc_2 = {
+      'length': [
+        {'meters': 'STANDARD'},
+        'feet'
+      ],
+      'level': 'length',
+    }
+    folder = unit_lib.UnitFolder(_GOOD_PATH)
+
+    folder.AddFromConfig([doc_1, doc_2], '{0}/file.yaml'.format(_GOOD_PATH))
+
+    self.assertIsInstance(folder.local_namespace.GetFindings()[0],
+                          findings_lib.DuplicateMeasurementAliasError)
+
 
 if __name__ == '__main__':
   absltest.main()
--- a/tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
+++ b/tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
@@ -28,17 +28,17 @@ _GOOD_PATH = '{0}/units/anyfolder'.format('mynamespace')
 
 class UnitLibTest(absltest.TestCase):
 
-  def testUnitUniverseGetUnitsMap(self):
+  def testUnitUniverseGetUnitsForMeasurement(self):
     folder = unit_lib.UnitFolder(_GOOD_PATH)
     namespace = folder.local_namespace
-    namespace.InsertUnit(unit_lib.Unit('degrees_celsius', 'temperature', False))
-    namespace.InsertUnit(unit_lib.Unit('kelvins', 'temperature', True))
+    namespace.InsertUnit('temperature', unit_lib.Unit('degrees_celsius', False))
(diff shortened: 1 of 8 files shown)


tools/validators/ontology_validator/yamlformat/validator/unit_lib.py
```diff
@@ -28,31 +29,49 @@
...
 
     Args:
       namespace: UnitNamespace
 
     Returns:
-      The units map from the namespace.
+      The namespace.
     """
-    return namespace.units
+    return namespace
```
kswang2400, line 53: i'm not seeing where this function is being used, but it seems pretty pointless now? can we just delete this?

tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
```diff
@@ -170,26 +173,74 @@ def testUnitFolderAddFromConfigTooManyStandardUnits(self):
...
+      'length': 'distance',
+      'level': 'length',
+    }
+    folder = unit_lib.UnitFolder(_GOOD_PATH)
+
+    folder.AddFromConfig([doc], '{0}/file.yaml'.format(_GOOD_PATH))
+
+    self.assertIsInstance(folder.local_namespace.GetFindings()[0],
+                          findings_lib.MeasurementAliasIsAliasedError)
+
```
kswang2400, line 243: add a test for DuplicateMeasurementAliasError?

```
    doc = {
      'distance': [
        {'meters': 'STANDARD'},
        'feet'
      ],
      'level': 'distance',
    }

    doc2 = {
      'length': [
        {'meters': 'STANDARD'},
        'feet'
      ],
      'level': 'length',
    }
```

kswang2400 (approved):
minor comments, LGTM

tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
```diff
@@ -170,26 +173,74 @@ def testUnitFolderAddFromConfigTooManyStandardUnits(self):
...
+    }
+    folder = unit_lib.UnitFolder(_GOOD_PATH)
+    expected_units = ['meters', 'feet']
+
+    folder.AddFromConfig([doc], '{0}/file.yaml'.format(_GOOD_PATH))
+    distance_units = folder.local_namespace.GetUnitsForMeasurement('distance')
+    length_units = folder.local_namespace.GetUnitsForMeasurement('length')
+
+    self.assertEmpty(folder.local_namespace.GetFindings())
+    self.assertCountEqual(expected_units, distance_units)
```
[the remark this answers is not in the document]
dplute: assertCountEqual actually does check that both sequences contain the elements, but they are not required to be in the same order.  It's just a weird name for what the method does.

tools/validators/ontology_validator/yamlformat/validator/presubmit_validate_types_lib.py
```diff
@@ -171,9 +156,6 @@ def GetUnitsForMeasurement(self,
         incremented.
     Returns: a string representing the unit or None if units are defined.
     """
-    if not self.unit_universe_reverse_map:
```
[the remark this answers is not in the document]
dplute: Added a check that returns if self.unit_universe is empty.

tools/validators/ontology_validator/yamlformat/validator/unit_lib.py
dplute: This is part of the way FindingsUniverse works and it's populating the _namespace_map that the other methods use.

tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
dplute: Added

tools/validators/ontology_validator/yamlformat/tests/unit_lib_test.py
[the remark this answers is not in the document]
kswang2400: got it, thanks for clarifying :)

tools/validators/ontology_validator/yamlformat/validator/unit_lib.py
kswang2400: ok, makes sense to leave it for now

kswang2400 (approved):

PR #422, opened 2021-11-25
Issue 260 units - part 2

pisuke:
@tsodorff @charbull @berkoben the previous units related PR (#276) diverged too much from master and I wasn't able to reconcile the code base, so I've started a new one.


pisuke:
@tsodorff @berkoben @charbull it looks like the [validator still fails when I declare the same units for two different measurements](https://github.com/google/digitalbuildings/runs/4355132718?check_suite_focus=true#step:5:17).
This still doesn't allow me to create measurements like `length` and `level` using the same unit `meter`.

mschulze17:
@dplute to review

dplute:
@pisuke Our solution to duplicate units was to allow specifying that a measurement uses the same list of units as another measurement, rather than directly defining the same units multiple times.  Please see the updated docs for the new syntax: https://github.com/google/digitalbuildings/blob/master/ontology/docs/ontology_config.md#units

pisuke:
Thank you @dplute I've updated the format for those measurements.
All checks now pass, @tasodorff can you PTAL?

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -208,29 +208,69 @@ descriptor:
...
   rate: "ratio of change compared to a step or time."
+  reactivepower: "Difference between the electricity supplied and the electricity converted into useful power."
   resistance: "Electrical resistance."
+  rotationalvelocity: "Vector measure of rotation rate, that refers to how fast an object rotates or revolves relative to another point."
+  specificenergy: "Energy per unit mass."
   specificenthalpy: "Measure of energy of air per unit mass."
   temperature: "Measures the temperature of media within a process or system."
-  time: "defined duration of something."
+  temperaturechangerate: "Difference in temperature per unit time."
+  temperaturedifference: "Difference of temperature."
```
tasodorff, line 267: is this a `differential_temperature`, where we would use the measurement descriptor `differential`?

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -208,29 +208,69 @@ descriptor:
...
+  temperatureintegral: "Integral of temperature over time."
```
tasodorff, line 268: would this better be described as a `temperature_accumulator`, similar to the way we have `energy_accumulator` for meters?

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -208,29 +208,69 @@ descriptor:
...
+  thermalconductivity: "Measure of a material's ability to conduct heat."
+  thermalcapacity: "Amount of heat to be supplied to an object to produce a unit change in its temperature."
+  thermaltransmittance: "Rate of transfer of heat through matter. "
```
tasodorff, line 271: we already have the precedent of `thermal_power_sensor` for heat transfer. is this the same thing?

ontology/yaml/resources/units/units.yaml
```diff
@@ -32,27 +45,83 @@ density:
...
+- megabtus
+- megajoules
 - megawatt_hours
+- therms
+- ton_hours
+- watt_hours
+energydensity:
+- kilojoules_per_kilogram: STANDARD
+entropy:
+- joules_per_degree_kelvin: STANDARD
```
tasodorff, line 81: nit: technically kelvin is not in degrees.

ontology/yaml/resources/units/units.yaml
```diff
@@ -32,27 +45,83 @@ density:
...
+- cycles_per_minute
+- kilohertz
+- megahertz
+- per_hour
+- per_minute
+- per_second
+- percent_per_second
 humidity:
 - percent_relative_humidity: STANDARD
+- grams_of_water_per_kilogram_dry_air
```
tasodorff, line 110: we model humidity as a percentage, not as this. we cover absolute humidity differently, since it can either be measured in units of mass/mass or mass/volume. see here: https://github.com/google/digitalbuildings/blob/407075cbfbd3d81e6c2ce0c25cc160e3238d9225/ontology/yaml/resources/fields/telemetry_fields.yaml#L999

ontology/yaml/resources/units/units.yaml
```diff
@@ -61,45 +130,145 @@ linearacceleration:
...
+- centimeters_of_mercury
+- centimeters_of_water
+- millimeters_of_mercury
 - inches_of_water
+- inches_of_mercury
 - pounds_force_per_square_inch
-- atmospheres
+- bars
+- millibars
+pressurechange:
```
tasodorff, line 201: what is this? not sure what this is.

ontology/yaml/resources/units/units.yaml
```diff
@@ -61,45 +130,145 @@ linearacceleration:
...
+- megavolt_amperes_reactive
 resistance:
 - ohms: STANDARD
 - kiloohms
+- megaohms
+- milliohms
+rotationalvelocity:
+- revolutions_per_minute: STANDARD
+specificenergy:
+- joules_per_kilogram_degree_kelvin: STANDARD
```
tasodorff, line 217: remove degree

ontology/yaml/resources/units/units.yaml
```diff
@@ -61,45 +130,145 @@ linearacceleration:
...
 - degrees_fahrenheit
+temperaturechangerate:
+- kelvins_per_second: STANDARD
+- kelvins_per_minute
+- kelvins_per_hour
+- degrees_celsius_per_hour
+- degrees_celsius_per_minute
+- degrees_fahrenheit_per_hour
+- degrees_fahrenheit_per_minute
+temperaturedifference:
```
tasodorff, line 236: `differential_temperature` already exists

ontology/yaml/resources/units/units.yaml
```diff
@@ -61,45 +130,145 @@ linearacceleration:
...
+- delta_degrees_kelvin: STANDARD
```
tasodorff, line 237: degrees goes away

ontology/yaml/resources/units/units.yaml
```diff
@@ -61,45 +130,145 @@ linearacceleration:
...
+- delta_degrees_celsius
+- delta_degrees_fahrenheit
+temperatureintegral:
```
tasodorff, line 240: see comment above

tasodorff (changes requested):
Please remove `degrees` from kelvin. Also, please remove from the PR anywhere that we have already defined a way to work with measurements and measurement descriptors, particularly for `differential_temperature`.

ontology/yaml/resources/subfields/subfields.yaml
pisuke: The term differential usually refers to the methodology of HVAC control (differential temperature control), while with this unit we just want to express the difference between two set points or between a set point and a sensor value.
We could also decide to just remove any "delta" / difference type of unit.
About the Kelvin unit, should we avoid the s for the plural following the [scientific recommentation](https://www.oxfordlearnersdictionaries.com/definition/english/kelvin)?

ontology/yaml/resources/subfields/subfields.yaml
pisuke: By having a specific measurement rather than a measurement descriptor we can introduce the concept of degree days, that is very important for simulation reasons.

ontology/yaml/resources/subfields/subfields.yaml
pisuke: I prefer to use standard scientific terms for measurements. Thermal power describes how fast heat is produced in an energy generation system, while the terms included here are more about thermal insulation and thermal inertia.

ontology/yaml/resources/units/units.yaml
pisuke: They are both adimensional units.
Maybe we should rename "humidity" to "relative humidity" and then add "absolute humidity"?

ontology/yaml/resources/units/units.yaml
pisuke: It is the unit [**atm**](https://en.wikipedia.org/wiki/Standard_atmosphere_(unit))

pisuke: I had actually removed it, but I can add it back.

ontology/yaml/resources/units/units.yaml
pisuke: Thanks for spotting this, done

ontology/yaml/resources/units/units.yaml
pisuke: Done

ontology/yaml/resources/units/units.yaml
pisuke: Done.

charbull:
Hi @pisuke can you take a look to resolve some conflicts please?

pisuke:
> Hi @pisuke can you take a look to resolve some conflicts please?

Done and tests passed, thanks for the nudge!

ontology/yaml/resources/subfields/subfields.yaml
tasodorff: the problem i have with integral is that it is imprecise about what you are integrating with respect to -- is it the integral of temperature with respect to distance? that would be an equally valid assumption, because there is no explicit callout for time. i suggest `timetemperature`, which is about the closest i can find to a proper name for the concept we are discussing: https://en.wikipedia.org/wiki/Time_temperature_indicator#cite_note-2

ontology/yaml/resources/units/units.yaml
tasodorff: Relative and absolute humidity are not both dimensionless. 

Relative humidity is a percentage, which is dimensionless. 

Absolute humidity is can be dimensionless (g water/kg dry air), but it can be represented in ways that have units (g water/m^3 dry air).

I think the easiest thing to do here is add `grams_of_water_per_kilogram` as units under concentration, and then we can use the fields from lines 999/1000 i linked above.

ontology/yaml/resources/subfields/subfields.yaml
tasodorff: The appropriate way to name deltas is with the measurement descriptor `differential` as has been set previously (as the units of measurement will ultimately be in terms of temperature, not some new unit). I agree, remove the delta.

Agree, the plural of kelvin will be kelvin.

ontology/yaml/resources/subfields/subfields.yaml
tasodorff: After deliberating on this, I think it most appropriate to name this as `timetemperature`, which is the closest I could find to a name for the concept we are discussing: https://en.wikipedia.org/wiki/Time_temperature_indicator

I prefer this to `temperatureintegral` because `integral` is not specific about what the integration is with respect to. You could, with equal validity, assign `temperatureintegral` units of `kelvin_meters` or `kelvin_kilograms`.

ontology/yaml/resources/units/units.yaml
tasodorff: i am confused. where is this used? what is this fundamental concept? i am not familiar with anything that is derived as pressure per unit temperature.

ontology/yaml/resources/subfields/subfields.yaml
tasodorff: My concern is that this breaks precedent, and that most of these concepts can easily be constructed in a way that doesn't change precedent or readability: `thermal_powertransmittance` is equally readable and allows us to reuse the concept of `powertransmittance` when it arises from other sources (such as radiative transmittance). 

Perhaps a compromise here is to simplify to `transmittance`, which appears to have consistent use. Then, if further qualification is needed, `thermal_transmittance` can be used, and our standard can be maintained.

tasodorff (changes requested):
Still need some fields changed/removed, and some units moved around. See comments.

ontology/yaml/resources/subfields/subfields.yaml
pisuke: OK, both `kelvin` and removal of `temperaturedifference` done.

ontology/yaml/resources/units/units.yaml
pisuke: I've decided to remove this unit for now.

ontology/yaml/resources/units/units.yaml
pisuke: `atmosphere` is a pressure unit, what is confusing?

ontology/yaml/resources/units/units.yaml
pisuke: `temperaturedifference` removed

ontology/yaml/resources/units/units.yaml
pisuke: done

ontology/yaml/resources/units/units.yaml
pisuke: So it is not about `atmosphere` and rather about `psi_per_degree_farenheit`?
If so, this comes directly from the BACnet standard. See [this example](https://store.chipkin.com/articles/bacnet-the-loop-object) of where it is used to define a PID loop.

tasodorff: No my question is more around what the phenomenon is we are measuring. For example, `power` is the rate of change of energy into a system; it is in units of force-distance over unit time. Almost all of the concepts you have defined here are named after the phenomenon they describe. This is the exception. 

It sounds like the application is for a PID controller? I assume it takes as input degrees fahrenheit and outputs PSI, and so this is simply an arbitrary conversion factor? 

If that is the case, I am having a hard time understanding the use case considering we do not model PID loops (they are way too low level). Is this really something we want to commit?

mschulze17: @pisuke nudge for the above conversation :)

pisuke:
@charbull @tasodorff This PR keeps on getting stale, could you please resolve conflicts, approve and merge?
Happy to resolve conflicts myself if you can merge it soon afterwards.

charbull:
@pisuke can you please resolve the conflicts? I can see it was not approved yet to merge.

mschulze17:
@tasodorff PTAL

tasodorff:
@pisuke the only thing I have concern over is the `pressurechange`. If this is just for PID loops then I would rather not include it, unless we came up with a better name for it. We can discuss this separately and come up with a good name for it, if you find the use-case to be compelling enough; for now, might be more expedient just to remove it, resolve the merge conflicts, and get it committed.

tasodorff (approved):
LGTM

PR #434, opened 2021-12-07
Updated ABSTRACT, AHU, CH, and MAU types.

tom-skoczylas:
Added new types to ABSTRACT, AHU, CH, and MAU. @tasodorff @tsodorff @mschulze17


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -3405,6 +3405,64 @@ AHU_ECON_EFSS2X_EFVSC2X_EPC_HT4SC_HTWHLSTC_SFC_SFSS2X_SFVSC2X_SSPC:
...
+  - DFSS
+  - DFVSC
+  - DSP
+  - DX4DC
+  - ECONZ
+  - EFSS
+  - ZSPC
+
+AHU_BSPC_DX4SC_EC2SC_ECON_EFSS2X_EFVSC2X_SARC_SFSS4X_SFVSC_SSPC:
+  description: "Multi-zone AHU."
```
tasodorff, line 3451: Add comment about evaporative precooling, which it looks like this does. Please also confirm that the `EC2SC` is for supply temp control, and that this isn't actually evaporative cooling for the condensing coils.

ontology/yaml/resources/HVAC/entity_types/MAU.yaml
```diff
@@ -568,6 +568,16 @@ MAU_DF2XSS_DFVSC2X_DSP_HTZC:
   - DSP
   - HTZC
 
+MAU_DFSS_DFVSC_HPDDC_ZTM:
+  description: "Dual discharge setpoint control with zone monitoring for heat pump MAU."
+  is_canonical: true
+  implements:
+  - MAU
+  - DFSS
+  - DFVSC
```
tasodorff, line 577: How does it do speed control, and why is that not part of this dataset?

tasodorff (changes requested):
A few comments regarding operation, and requested update to description.

ontology/yaml/resources/HVAC/entity_types/MAU.yaml
tom-skoczylas: It has linked speed control with a Melink KEF on the same control program. Not sure what you mean about it not part of this dataset.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
tom-skoczylas: It appears to be supply controlled. This is for US-MTV-1015 AC-1.

ontology/yaml/resources/HVAC/entity_types/MAU.yaml
tasodorff: ah, so the melink sends it a speed setpoint to run at.

tasodorff (approved):

PR #445, opened 2022-01-10

rniemo-g:
PiperOrigin-RevId: 420800152


charbull:
can you please add your github user name internally, go/cla

charbull:
I see what is happening, github does not work like our internal tooling. 
The proposed pr is deleting all the files that are not in the `ontology/**`

PR #456, opened 2022-01-18
Commits for typical project atmosphere sensors

tasodorff:
New project atmosphere types.


ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,62 @@ SENSOR_FDPM:
...
+  description: "Atmosphere formaldehyde combined sensor with local temp and RH readings."
+  is_canonical: true
+  implements:
+  - /SENSOR
+  uses:
+  - zone_air_formaldehyde_concentration_sensor
+  - local_air_relative_humidity_sensor
+  - local_air_temperature_sensor
+
+SENSOR_OZONE_NO_ATM:
```
berkoben, line 146: why not NO2 if that's what we're measuring?

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,62 @@ SENSOR_FDPM:
...
+  - local_air_relative_humidity_sensor
+  - local_air_temperature_sensor
+
+SENSOR_PM_ATM:
+  description: "Atmosphere PM 0.5, PM 2.5, etc. combined sensor with local temp and RH readings."
+  is_canonical: true
+  implements:
+  - /SENSOR
+  uses:
+  - thirtysecondrolling_average_particle_diameter_sensor
```
berkoben, line 166: Why do all the others have zone and this one not?

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,62 @@ SENSOR_FDPM:
...
+  - zone_air_relative_humidity_sensor
```
berkoben, line 167: more thoughts about "zone":  the case for this prefix was pretty clear-cut when strictly talking about HVAC equipment, but when we're talking about a standalone sensor, the expectations for where it is placed become less clear.  I'm not sure we 100% solve this in the current PR, but perhaps we think ahead to what the ultimate solution is and see how it informs the current PR.  My first idea:
```
SPECIFIC_SENSOR:
  implements:
  - ABSTRACT_SENSOR:
    air_temperature_sensor:zone_air_temperature_sensor
```
The idea above being that you could remap fields to be more specific when inheriting a type, allowing us to not overspecify the sensor to start out with.

berkoben (changes requested):

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,62 @@ SENSOR_FDPM:
...
+SENSOR_CO_CO2_ATM:
+  description: "Atmosphere CO and CO2 combined sensor with local temp and RH readings."
+  is_canonical: true
+  implements:
+  - /SENSOR
+  uses:
+  - thirtysecondrolling_average_zone_air_co_concentration_sensor
+  - zone_air_co_concentration_sensor
+  - zone_air_co2_concentration_sensor
+  - local_air_relative_humidity_sensor
```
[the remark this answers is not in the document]
tasodorff: these are sensor clusters, and this particular RH and temp sensor are located right on the sensor to detect conditions at the CO/CO2 sensor heads. I suppose technically they can be zone temp sensors (in the same way that the CO and CO2 sensors are techincally measuring the space they are in), but they aren't measuring the zone so much as they are measuring the conditions at the sensor itself for calibration purposes.

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,62 @@ SENSOR_FDPM:
...
+  - local_air_temperature_sensor
+
+SENSOR_FORM_ATM:
```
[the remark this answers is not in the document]
tasodorff: I can live with HCHO.

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
tasodorff: oops. forgot to add the 2.

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,62 @@ SENSOR_FDPM:
     - local_air_temperature_sensor
     - battery_charge_percentage_sensor
     - signal_strength_percentage_sensor
+
+SENSOR_CO_CO2_ATM:
```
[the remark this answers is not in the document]
tasodorff: i used ATM as a callout for the project atmosphere sensors, but in retrospect its unnecessary. ive removed it

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
tasodorff: I like this idea, but don't think we are going to solve it here. I will add a spike to think about this one.

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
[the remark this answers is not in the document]
berkoben: Would internal be more appropriate in this case?

berkoben (changes requested):

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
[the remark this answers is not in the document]
drewdepriest: Instead of Local/Internal - what about "ROOM" (already defined as part of FACILITIES entity) or "SPACE"?

**Control vs Monitor**
From a FacOps system perspective, there exists a distinction between "zone" and "room/space" where "zone" implies an element of _control_ (i.e. HVAC terminal unit, lighting ballast, etc.) and room/space implies _measure_ (i.e. occ sensors for space utilization analysis, IAQ/IEQ sensors, light intensity, sound levels, etc). 

When I want to analyze data from _monitor_ devices, I want to know their relation to space (and a unique spaceId). For _control_ devices, that analysis generally flows upstream from the "zone" through various air and hydronic systems. In a future digital twin state, we'll want to map zone to space, as they won't always be 1:1 (i.e. a single tstat that serves multiple collab spaces).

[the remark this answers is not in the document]
tasodorff: i think for simplicity we will rename them from local to `zone`; after discussing with Keith, we went down the rabbit hole of "well, technically, all sensors measure their temperatures 'locally'" and so for consistency we simply set reverted the `local` descriptor to `zone`. 

From the perspective of DBO, a zone is simply a defined region within a building; not necessarily a single room, it can be a collection of rooms or perhaps a subregion of a single room. The question of which rooms map to which zones can be handled elsewhere, particularly through connections: this zone connects to these rooms. The semantic distinction between `zone` versus `room` are legitimate, and we do have `room` as a defined subfield, but in practice it would necessitate that the modeler *know* that the sensor (or device its connected to) serves only a single room; zone is less specific, and therefore allows the modeler to assign it without having to care about whether it serves a single room or multiple.

To summarize: we will remove `local` and use `zone` instead; we will avoid `room` so that we don't have to care about whether the sensor really does apply to just a single room or not.

tasodorff:
@berkoben please review; all your changes are belong to us

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,56 @@ SENSOR_FDPM:
...
+  - zone_air_formaldehyde_concentration_sensor
+
+SENSOR_ZA_OZONE_NO2:
+  description: "Atmosphere ozone and NO2 combined sensor with local temp and RH readings."
+  is_canonical: true
+  implements:
+  - /SENSOR
+  - ZA
+  uses:
+  - zone_air_no2_ozone_concentration_sensor_1
```
berkoben, line 151: Thought: maybe we make the internal sub-sensors optional?  That might make this type more reusable
zone_air_no2_ozone_concentration_sensor_1
zone_air_no2_ozone_concentration_sensor_2
zone_air_no2_probability_percentage_sensor

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1066,3 +1066,10 @@ literals:
 
 - min_zone_air_relative_humidity_sensor
 - min_zone_air_relative_humidity_setpoint
+
+## New fields for air quality sensors
+- thirtysecondrolling_average_zone_air_co_concentration_sensor
+- thirtysecondrolling_average_zone_air_particle_diameter_sensor
+- zone_air_no2_probability_percentage_sensor
```
berkoben, line 1073: and here.  I note that this one doesn't show the internal NO2 sensors.  Is that an omission or are these devices just different?

berkoben (changes requested):

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
tasodorff: accepted

ontology/yaml/resources/fields/telemetry_fields.yaml
tasodorff: this is the `fields.yaml` file, not a type definition. is this comment regarding the field itself?

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -56,6 +57,7 @@ component:
   wheel: "Component used for transfer of heat from incoming to outgoing air streams. "
   winding: "Conductive wires wrapped around a magnetic core."
 descriptor: 
+  adjusted: "A descriptor indicating the value has been changed from the raw reading."
```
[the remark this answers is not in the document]
tasodorff: removed

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -120,3 +120,57 @@ SENSOR_FDPM:
...
+  - /SENSOR
+  - ZA
+  uses:
+  - zone_air_ozone_concentration_sensor
+  - zone_air_voc_concentration_sensor
+  - zone_air_no2_concentration_sensor
+  opt_uses:
+  - zone_air_no2_ozone_concentration_sensor_1 #Note these sensors cannot distinguish between NO2 and O3
+  - zone_air_no2_ozone_concentration_sensor_2
+  - zone_air_no2_probability_percentage_sensor #indicates the probability that the combination of sensors is sensing NO2
```
berkoben, line 157: nit: if the comment wraps, put it on its own line above the relevant config line.

ontology/yaml/resources/units/units.yaml
```diff
@@ -22,11 +22,13 @@ concentration:
 - parts_per_million
 - parts_per_billion
 - grams_per_kilogram # Can convert approximately using Avogadro
+- parts_per_cubic_centimeter # can be approximated using Avogadro
```
berkoben, line 24: This field and the field above it are two different types of concentration (volumentric and mass) that aren't trivially convertible to ppm or ppb.  They both need their own measurement subfields.

berkoben (changes requested):

berkoben (approved):

PR #459, opened 2022-01-21

kswang2400:
PiperOrigin-RevId: 423126988


charbull (commented):
It deletes the scoring code.

trav3711 (changes requested):
Seems like you might need to sync back to g3 again for the scoring code.

PR #462, opened 2022-01-24
ParseConfig class + initialization tests

ahemphill:
Addressed comments in previous PR. I've made a task to revisit the `append_types()` test once the output format, etc. firms up.


tools/scoring/tests/fixtures/files/proposed/building_config_example.yaml
```diff
@@ -0,0 +1,180 @@
+# Copyright 2022 Google LLC
```
charbull, line 1: can you please put the files in 
```
tests/samples/proposed/
tests/samples/solution/
```

Just to be compatible with the other tools.

tools/scoring/score/parse_config.py
```diff
@@ -0,0 +1,104 @@
...
+                    source.translation[target_field] = source.translation[
+                        source_field]
+                    source.translation[
+                        target_field].std_field_name = target_field
+                    del source.translation[source_field]
+                  except KeyError:
+                    translations_absent.append(
+                        f'{link.source}.translation.{source_field}')
+
+      print(f'{file_type} translations absent: ' +
```
charbull, line 99: instead of printing here,

we can print but also return those elements and then check the returned content in the tests below.
What do you think?

charbull (changes requested):
minor minor :)

tools/scoring/score/parse_config.py
ahemphill: Tested the actual calls. Created a task to revisit when the rest firms up.

tools/scoring/tests/parse_config_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+    self.assertEqual(type(self.parse.parsed['proposed']),
+                     dict)  # Dict[str, EntityInstance]
+    self.assertEqual(type(self.parse.parsed['solution']),
+                     dict)  # Dict[str, EntityInstance]
+
+    self.assertEqual(type(self.parse.scores), dict)
+
+  @patch('builtins.print')
+  def testAppendTypes(self, mock_print):
```
charbull, line 49: this is dirty but ok since you have a todo above to revisit it

tools/scoring/tests/parse_config_test.py
```diff
@@ -0,0 +1,65 @@
...
+    self.assertEqual(mock_print.call_count, 4)
+    calls = [
+        call('proposed translations absent: 0 (from 0 links)'),
+        call('proposed types absent: 0 (0 instances)'),
+        call('solution translations absent: 0 (from 0 links)'),
+        call('solution types absent: 0 (0 instances)')
+    ]
+    mock_print.assert_has_calls(calls)
+
+  # def testMatchReportingEntities(self):
```
charbull, line 60: remove

tools/scoring/score/parse_config.py
charbull: can you add a todo please.
In the future, lets return variables it is cleaner than checking prints :)

tools/scoring/score/parse_config.py
```diff
@@ -0,0 +1,104 @@
...
+                source.type = self.universe.GetEntityType(
+                    source.namespace, source.type_name)
+                if source.type is None:
+                  types_absent.append(getattr(source, type_or_name))
+              link.source_type = source.type
+              for target_field, source_field in link.field_map.items():
+                if target_field != source_field:
+                  try:
+                    source.translation[target_field] = source.translation[
+                        source_field]
```
charbull, line 91: this can be broken down into small functions, in the future PR while you revisit the returns instead of prints.
please add todo

charbull (approved):

PR #464, opened 2022-01-28
Match reporting entities + tests

ahemphill:


tools/scoring/score/parse_config.py
```diff
@@ -103,3 +104,26 @@ def append_types(self):
...
+      Matches reporting entities by `cloud_device_id`
+
+      Returns:
+        List of paired proposed and solution entities
+    """
+    matches = []
+    for solution_entity in solution.values():
+      if solution_entity.cloud_device_id is None:
+        continue  # as this is not a reporting device
+      match = None
```
charbull, line 123: nit: matched_entity 
just to be consistent?

tools/scoring/tests/fixtures/files/proposed/match_reporting_entities.yaml
```diff
@@ -0,0 +1,54 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file is used as an example of a valid building configuration with
+# the most common and important features, e.g. translations and links.
+
+# This WILL match as cloud_device_id corresponds and translation is present
```
charbull, line 18: can we re-use the existing files or consolidate?

the path needs to be reworked to be in here: https://github.com/google/digitalbuildings/tree/master/tools/scoring/tests/samples

tools/scoring/tests/parse_config_test.py
```diff
@@ -57,6 +60,28 @@ def testAppendTypes(self, mock_print):
     ]
     mock_print.assert_has_calls(calls)
 
+  def testMatchReportingEntities(self):
+    proposed = validator.Deserialize(
+        ['tests/fixtures/files/proposed/match_reporting_entities.yaml'])[0]
```
charbull, line 65: https://github.com/google/digitalbuildings/blob/master/tools/scoring/tests/samples/proposed/building_config_example.yaml

tools/scoring/tests/parse_config_test.py
```diff
@@ -57,6 +60,28 @@ def testAppendTypes(self, mock_print):
...
+    solution = validator.Deserialize(
+        ['tests/fixtures/files/solution/match_reporting_entities.yaml'])[0]
```
charbull, line 67: https://github.com/google/digitalbuildings/blob/master/tools/scoring/tests/samples/solution/building_config_example.yaml

tools/scoring/tests/parse_config_test.py
```diff
@@ -57,6 +60,28 @@ def testAppendTypes(self, mock_print):
...
+
+    matches = parse_config.ParseConfig.match_reporting_entities(
+        proposed=proposed, solution=solution)
+
+    self.assertEqual(len(proposed), 4)
+    self.assertEqual(len(solution), 4)
+    self.assertEqual(len(matches),
+                     2)  # number of valid reporting entities in solution
```
charbull, line 75: I can see only one match here: `2599571827844401` which one is the second match?

charbull (changes requested):

tools/scoring/tests/fixtures/files/proposed/match_reporting_entities.yaml
ahemphill: The existing files could be used but in that circumstance it's much more difficult to discern what is being tested.

tools/scoring/tests/parse_config_test.py
ahemphill: As above; I think it's logically easier to map what is being tested to the more concise file.

tools/scoring/tests/parse_config_test.py
ahemphill: "matches" is a bit of a misnomer as it's a list of valid reporting entities and their matches _if they exist_, hence the comment. Would "pairs" be more intuitive?

ahemphill: Further thoughts: should this just be a dict with IDs to be referenced against the original "proposed" and "solution" dicts? It would mean two additional lookups for each pair but the signature would be much clearer.

charbull (approved):

PR #467, opened 2022-02-01

kswang2400:
PiperOrigin-RevId: 425717115


.github/workflows/python_tools.yml
```diff
@@ -51,18 +51,19 @@ jobs:
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/explorer
-    #---------- Configuration scoring ----------#
```
charbull, line 54: those are deleted

charbull (changes requested):
it is deleting the scoring code.
I think you can update the copybara file to ignore the scorer for now?

.github/workflows/python_tools.yml
kswang2400: Done

trav3711 (approved):
I'm not sure about the scoring code but all of the changes I want are in there.

charbull:
closing it in attempt to sync

PR #468, opened 2022-02-01
Wrappers for dimension invocation + tests

ahemphill:


tools/scoring/score/parse_config.py
```diff
@@ -177,3 +177,57 @@ def retrieve_reporting_translations(
       }
 
     return translations
+
+  @staticmethod
+  def aggregate_scores_nondbo(
+      # TODO: create Dimension and Score types to replace generics
```
charbull, line 183: can you please create those first, it will help me understand what they are :)

charbull (changes requested):

tools/scoring/score/parse_config.py
ahemphill: https://github.com/google/digitalbuildings/pull/468/commits/cfe49ce45c291cf1421018d3baf5962616179d38

ahemphill: The `Dimension` type(s) won't add much beyond reference convenience as they each only take two arguments; I'd like to punt on those until at least one piece of the core functionality is landed.

tools/scoring/score/dimensions/result.py
```diff
@@ -0,0 +1,83 @@
...
+    Container for floating-point results which
+    quantify success in each dimension
+
+    Attributes:
+      correct_virtual: Number of correct attempts within virtual devices
+      correct_reporting: Number of correct attempts within reporting devices
+      correct_ceiling_virtual: Number of correct attempts possible
+        within virtual devices
+      correct_ceiling_reporting: Number of correct attempts possible
+        within reporting devices
```
charbull, line 30: not sure I understand the difference between the `ceiling` ones and non ceiling.

tools/scoring/score/dimensions/result.py
```diff
@@ -0,0 +1,83 @@
...
+      incorrect_virtual: Number of incorrect attempts within virtual devices
+      incorrect_reporting: Number of incorrect attempts within reporting devices
+
+    Properties:
+      result_composite: Calculated result for all devices
```
charbull, line 35: total_result?

tools/scoring/score/parse_config.py
```diff
@@ -177,3 +178,57 @@ def retrieve_reporting_translations(
...
+
+  @staticmethod
+  def aggregate_results_nondbo(
+      # TODO: create Dimension type to replace generic
+      *,
+      dimensions: List[Callable[[Dict], Dict]],
+      translations: Dict[str, List[Tuple[str, Any]]]) -> Dict[str, Result]:
+    """
+      Wrapper which outputs a dictionary of results by invoking
+      each specified `Dimension` with the `translations` argument
```
charbull, line 190: what is a dimension? the `Result` class is a dimension?

tools/scoring/score/parse_config.py
```diff
@@ -177,3 +178,57 @@ def retrieve_reporting_translations(
...
+        Dictionary with dimension names as keys and `Result`s as values
+    """
+    results = {}
+    for dimension in dimensions:
+      # Invoke the function and append the dictionary with its return value
+      results[dimension.__name__] = dimension(translations=translations)
+    return results
+
+  @staticmethod
+  def aggregate_results_dbo(
```
charbull, line 207: aggregate_results_by_dimension?
Are we calculating here the result by Dimension?

What is a dimension?
reporting devices
field matching etc all those are a dimension each?

it is a sort of a measurement?

charbull (commented):
have questions :)

tools/scoring/score/dimensions/result.py
ahemphill: `correct_{category}` is the number of attempts achieved whereas `correct_ceiling_{category}` is the number of attempts it is _possible_ to achieve. It's used in `result_{category}()` to normalize correct versus incorrect attempts.

tools/scoring/score/dimensions/result.py
ahemphill: More intuitive verbiage might be helpful but total implies a sum, I think…

tools/scoring/score/parse_config.py
ahemphill: A dimension is a metric which is being evaluated for the proposed file (i.e. the core functionality). I'd planned to have them separate but maybe that's unnecessary…

tools/scoring/score/parse_config.py
ahemphill: The suffix is for differentiating between the two types of dimension: less complex (which operate on the translations of each matched reporting pair) and more (which operate on the entire files). The categorization is almost meaningless save for the common signatures. You figured it out: a dimension is a measurement, and these methods just wrap every dimension for modularity.

tools/scoring/score/dimensions/result.py
charbull: what do we mean by it is `possible` to achieve? can you please give an example?

tools/scoring/score/dimensions/result.py
charbull: I see, sgtm

charbull (changes requested):
help me understand :)

tools/scoring/score/dimensions/result.py
ahemphill: "The test has four questions. It is possible to get four questions correct. The ceiling is four."

It's a little confusing because in the context of `correct`, "attempt" implies action taken. However, it's possible to get something wrong because no attempt was made.

"The ceiling is four. You attempted three and got two right. Despite only one incorrect guess, two were wrong: one by action and one by default."

charbull (approved):

PR #478, opened 2022-02-04
First core component; refinement of Dimension class

ahemphill:


tools/scoring/score/dimensions/dimension.py
```diff
@@ -11,71 +11,103 @@
...
     Container for floating-point results which
     quantify success in each dimension
 
     Attributes:
-      correct_virtual: Number of attempts achieved within virtual devices
-      correct_reporting: Number of attempts achieved within reporting devices
-      correct_ceiling_virtual: Number of attempts possible to achieve
+      translations: Proposed and solution translations
+        for all matched reporting entities. Assigned via argument
+      deserialized_files: Parsed configuration files.  Assigned via argument
```
charbull, line 27: what is a deserialized file ? do you have a list of entity instance?

tools/scoring/score/dimensions/dimension.py
```diff
@@ -11,71 +11,103 @@
...
+    self.correct_ceiling_reporting: int = None
+    self.incorrect_virtual: int = None
+    self.incorrect_reporting: int = None
 
-  def correct(self) -> int:
-    """ Number of attempts achieved within all devices """
-    return self.correct_virtual + self.correct_reporting
+    if not translations and not deserialized_files:
+      raise Exception(
+          '`translations` xor `deserialized_files` argument is required')
```
charbull, line 60: why do we need both? to look up virtual entities?

tools/scoring/score/dimensions/field_selection.py
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License.
+""" Core component """
+
+from typing import Any, Dict, List, Tuple
+
+from score.dimensions.dimension import Dimension
+
+
+class FieldSelection(Dimension):
+  """ Quantifies whether the correct fields were mapped (versus ignored) """
```
charbull, line 22: can you please add an example in the docstring what is a field?
is it the raw field? or the standard field from the ontology ?
What is a selection? it is present in the proposed?

charbull (changes requested):
questions & more comments :)

tools/scoring/score/dimensions/dimension.py
ahemphill: Exactly that: `deserialized_files` is the repackaged output of `validator.Deserialize()` (parse_config.py:55). I added more complete type annotation to `__init__()` definition below this. (I was holding off since I'm focusing on the less complex dimensions ATM.)

tools/scoring/score/dimensions/dimension.py
ahemphill: Correct; the simple dimensions receive `translations` and the complex dimensions receive `deserialized_files`. (There's likely to be some repetition in the virtual entity matching algorithm(s) which will have to be combined down the line.)

tools/scoring/score/dimensions/field_selection.py
ahemphill: Yes, successful "field selection" means "the raw fields were present in the proposed file."

tools/scoring/score/dimensions/dimension.py
charbull: can you add more details in the docstring about its content what it is inside it what kind of objects it has in it.
otherwise deseralized is too vague

tools/scoring/score/dimensions/dimension.py
```diff
@@ -11,71 +11,104 @@
...
-  correct_virtual: int
-  correct_reporting: int
-  correct_ceiling_virtual: int
-  correct_ceiling_reporting: int
-  incorrect_virtual: int
-  incorrect_reporting: int
+  def __init__(self,
+               *,
+               translations: Dict[str, List[Tuple[str, Any]]] = None,
+               deserialized_files: Dict[str, Dict[str, EntityInstance]] = None):
```
charbull, line 47: what are those `str` keys can you please add more info in the docstring. for both the translation and the deserialized files

tools/scoring/score/dimensions/dimension.py
charbull: can you please leave a comment why you need both

tools/scoring/score/dimensions/field_selection.py
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" Core component """
+
+from typing import Any, Dict, List, Tuple
+
+from score.dimensions.dimension import Dimension
+
+
+class FieldSelection(Dimension):
```
charbull, line 21: should we rename to `RawFieldSelection`?

charbull (changes requested):
need for comments please :)

tools/scoring/score/dimensions/dimension.py
ahemphill: They're `cloud_device_id`s… I'll make types.

tools/scoring/score/dimensions/dimension.py
ahemphill: Added some detail but I'm not sure how to be a lot more specific without poorly rehashing what will be specified with the type.

tools/scoring/score/dimensions/dimension.py
charbull: Dict[str, Dict[str, EntityInstance]]

there are two str?

charbull (changes requested):

ahemphill:
Yes: the first is the file type and the second is the entity ID. Adding types.

charbull (approved):

PR #484, opened 2022-02-07
Add new devices Air Ionizer + Pre-action fire alarm system + Solenoid valve

josephedwardchang:
new devices:

HVAC/ADY.yaml - Air Ionizer
SAFETY/FACP.yaml - Pre-action fire alarm system
SAFETY/LDS.yaml - Solenoid valve

some edits carried over from approved pull-request #386 (now closed) but somehow was not merged yet. I closed it (#386) thinking it was merged already by others as I don't have proper credentials to do the merge.


ontology/yaml/resources/HVAC/entity_types/ADY.yaml
```diff
@@ -29,3 +29,16 @@ ADY_BS_SS:
   - control_mode
   - schedule_run_command
   - control_status
+
+AION_BS_SS:
```
tasodorff, line 33: Is this a stand-alone piece of equipment? If so, it will need to be defined with a new general type (AION may work) but you have defined it here in the ADY.yaml file. This would need to be moved into its own file called AION.yaml.

If this isn't stand-alone (I assume this is probably the ionizer in an AHU) then it should be defined as part of that equipment. You might suggest something like: 
- `ionizing_filter_run_command`
- `ionizing_filter_run_status`

and then create a separate `IFSS` abstract type for ionizing filter start/stop.

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
```diff
@@ -29,3 +29,16 @@ ADY_BS_SS:
...
+  description: "Basic air ionizer with trip alarm, switch control/status, scheduled run, filter alarm."
+  is_canonical: true
+  implements:
+  - AION
+  - SS
+  opt_uses:
+  - lost_power_alarm
+  - control_mode
+  - schedule_run_command
```
tasodorff, line 42: If you intend this to be an abstract type (you have defined it in this PR) then you should require it. If you do have an ionizer that doesnt have a schedule, you should have two separate canonical types -- one with and one without that abstract type.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -571,4 +571,14 @@ FAN_SS_AL:
...
-  - control_status
\ No newline at end of file
+  - control_status
+
+FAN_SSSCFASR:
+  description: "Basic Exhaust fan with start/stop, speed command, fail alarm, scheduled run"
+  uses:
+  - exhaust_fan_speed_percentage_command
+  implements:
+  - CONTROL
```
tasodorff, line 581: remove; this goes with abstract types only.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -571,4 +571,14 @@ FAN_SS_AL:
...
   - run_mode
   - schedule_run_command
-  - control_status
\ No newline at end of file
+  - control_status
+
+FAN_SSSCFASR:
+  description: "Basic Exhaust fan with start/stop, speed command, fail alarm, scheduled run"
+  uses:
+  - exhaust_fan_speed_percentage_command
```
tasodorff, line 579: Shouldn't this be `FAN_SS_SRC`? And, if so, wouldn't it match the type directly above it (`FAN_SS_AL`)?

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -463,4 +463,15 @@ VAV_RD_CO2C:
...
\ No newline at end of file
+  - CO2C
+
+VAV_SD_SDM_CO2C_SRC:
+  description: "CO2 controlled VAV, with damper, flowrate and scheduled run timer"
+  is_canonical: true
+  implements:
+  - VAV
+  - SD
+  - SDM
```
tasodorff, line 474: Is there a separate supply damper that returns just a status and command? The combination of `SD` and `SDM` imply that all of these fields exist on the same device:
- `supply_air_damper_percentage_command`
- `supply_air_damper_command`
- `supply_air_damper_status`

If you have two separate supply dampers, then we would need to distinguish them from each other (`supply_air_damper_percentage_command` and `supply_air_isolation_damper_command`); if they are the same damper, how can it have both a percentage command and a binary command? We shouldn't have two commands for the same thing.

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
```diff
@@ -138,3 +138,25 @@ LC:
...
+
+ZLC:
+  description: "Lighting Control Zone light controls"
+  is_abstract: true
+  uses:
+  - zone_lighting_switch_command
+  - zone_lighting_switch_status
+  - zone_brightness_percentage_command
+
+LFA:
```
tasodorff, line 157: Is this alarm needed? If so, what does a switch alarm mean? What happens to cause this alarm?

ontology/yaml/resources/METERS/entity_types/WM.yaml
```diff
@@ -20,3 +20,12 @@ WM_STANDARD:
...
   - water_volume_accumulator
+
+WM_WFR:
+  description: "Standard building water meter with flow rate meter."
+  is_canonical: true
+  implements:
+  - WM_STANDARD
+  uses:
+  - water_volume_accumulator
+  - supply_water_flowrate_sensor
```
tasodorff, line 31: remove `supply` so it just reads `water_flowrate_sensor`. No need for the descriptor if it is not to differentiate it from something else. Will need to be defined (I don't think it already exists).

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1079,3 +1080,31 @@ literals:
...
+  - ON
+  - OFF
+- zone_lighting_switch_status:
+  - ON
+  - OFF
+- zone_lighting_switch_command:
+  - ON
+  - OFF
+- zone_brightness_percentage_command
+- lighting_switch_alarm:
```
tasodorff, line 1094: see comment above -- if this is needed, we should be more descriptive with what it does.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1079,3 +1080,31 @@ literals:
...
+- water_leak_cable_distance_sensor
+
+#added for new Pre-action Fire Supression system
+- supervisor_alarm:
+  - ACTIVE
+  - INACTIVE
+- water_flowrate_status:
+  - ACTIVE
+  - INACTIVE
+- air_pressure_status:
```
tasodorff, line 1108: I did not see this used anywhere else in the PR. Is this needed?

tasodorff (changes requested):
Please see the comments and provide some clarification.

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
josephedwardchang: on checking, it is stand-alone. I'll update again with your suggestions. thanks sir

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
josephedwardchang: thanks sir. I like it to be abstract and define 2 canonical ones for maximum flexibility.
Is it ok to make the 2 canonical types from abstract (one with schedule and one without) even if I only know and use the one with schedule? 
As of now, I don't know if there is ionizer without schedule. Do I just make 1 canonical device (with schedule) that we use and let others define the other later (without schedule) for their own use?

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: noted sir

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: hi sir. it is used in SAFETY/entity_types/FACP.yaml for (PAFAS:)

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: upon checking the device here. it seems to be a general fault alarm; i will use "failed_alarm" instead of defining this. thanks sir.

ontology/yaml/resources/METERS/entity_types/WM.yaml
josephedwardchang: noted sir

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
josephedwardchang: upon checking device, it appears to be a general fault alarm. I will use "failed_alarm" instead.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: i see. thanks sir. i will check the device again.

tasodorff:
@josephedwardchang i think you may have added a lot of new files to your PR accidentally. you had 15 files open for review before, but 122 now... and most of them are code. please go check that the files in your PR are the ones that you are modifying -- we can deal with branch conflicts together at the end. 

i will be providing feedback to your comments in those files that we've already begun discussing

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
tasodorff: it is okay to suggest both if you think it likely you will run into it in the field. something like this:
```
AION_SS:
  implements:
  - AION
  - SS
  opt_uses:
  - lost_power_alarm
  - control_mode

AION_SS_SRC:
  implements:
  - AION
  - SS
  - SRC
  opt_uses:
  - lost_power_alarm
  - control_mode
```

if you wanted to omit the one without schedule we can also do that. it might be easiest just to have both. i am okay with either one.

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
tasodorff: out of curiosity, where does this device live? is it in a duct (separate from the AHU)? does it stand in a space on its own? how does it fit into the system?

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
tasodorff: one more thing: please note the names for each type above is constructed from the implemented abstract types. you will find some that do not follow this convention, but in general all canonical types should try to follow that convention. just FYI in case you make more canonical types.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: we have documents in the repository which could be helpful. if you haven't read them already, please check these out:

- https://github.com/google/digitalbuildings/blob/master/ontology/docs/faq.md
- https://github.com/google/digitalbuildings/blob/master/ontology/docs/hvac_ahu.md
- https://github.com/google/digitalbuildings/blob/master/ontology/docs/hvac_fcu.md
- https://github.com/google/digitalbuildings/blob/master/ontology/docs/model_hvac.md

we are also working on some training modules which we will publish in the near future, but the docs above are your best resource for your current task.

so far your proposals have been good. I feel like you have a good understanding of the basic points (e.g. create abstract types when there is some functionality we want to reuse). most of my comments are more nuanced, and you will simply need some more time with the ontology before it becomes clearer to you. hopefully those docs above also clarify a few additional things, such as which fields we care about.

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
tasodorff: yes that would make it more clear. please adjust. and, if there are some common alarms that apply to a lot of devices, consider it as an abstract type of its own (that is, group multiple alarms together).

ontology/yaml/resources/fields/telemetry_fields.yaml
tasodorff: that sounds good. please delete this and update the types which use it.

ontology/yaml/resources/fields/telemetry_fields.yaml
tasodorff: ah i see it. thank you for clarifying.

tasodorff:
@josephedwardchang looking at this I still see many requested changes unresolved. can you please check that you have pushed your full set of changes?

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: Hi sir, sorry for late reply. I have an extra point needed that differs from AL:
```
uses:
  - exhaust_fan_speed_percentage_command
```

I also don't need some in the opt_uses, so I decided to make a new entry, i think it should be like this now:
```
FAN_SS_SPS_FA_SR:
  description: "Basic Exhaust fan with start/stop, speed sensor, fail alarm, scheduled run"
  uses:
  - exhaust_fan_speed_percentage_sensor
  - lost_power_alarm
  implements:
  - FAN
  - SS
  - SRC
```
Is it better to create a new type  for my specific use like above?
Or is it better to re-use exisiting types like FAN_AL (has some points more than is needed) and just apply MISSING on some points later, like below maybe?
```
FAN_SS_AL_SPS_SR:
  description: "Basic Exhaust fan with start/stop, fail alarm, speed sensor, scheduled run"
  uses:
  - exhaust_fan_speed_percentage_sensor
  implements:
  - FAN_SS_AL
  - SRC
```

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: > Is there a separate supply damper that returns just a status and command? The combination of `SD` and `SDM` imply that all of these fields exist on the same device:
> 
> * `supply_air_damper_percentage_command`
> * `supply_air_damper_command`
> * `supply_air_damper_status`
> 
> If you have two separate supply dampers, then we would need to distinguish them from each other (`supply_air_damper_percentage_command` and `supply_air_isolation_damper_command`); if they are the same damper, how can it have both a percentage command and a binary command? We shouldn't have two commands for the same thing.

For this sir, you are correct, I intend to apply a MISSING in the points later (supply_air_damper_percentage_command: MISSING). Because the SD and SDM combo is the nearest one I could find among all the others. I felt this is easiest to do on my part since I have minimal changes...but is that correct? Should I redo it like below and get only the parts I need? (Note: this is almost the same question earlier for FAN_SS_AL)

```
VAV_SDM_CO2C_SRC_HLFSS:
  description: "CO2 controlled VAV, with damper, hi/low flowrate setpoint and sensor and schedule run"
  is_canonical: true
  implements:
  - VAV
  - SDM
  - CO2C
  - SRC
  uses:
  - low_limit_supply_air_flowrate_setpoint
  - high_limit_supply_air_flowrate_setpoint
  - supply_air_flowrate_sensor 
```

josephedwardchang:
> @josephedwardchang looking at this I still see many requested changes unresolved. can you please check that you have pushed your full set of changes?

hi sir, sorry for delay. here are the updates, please check. I've also adjusted the files changed to the correct number by resetting my git repo to the DBO repo. Thanks

ontology/yaml/resources/HVAC/entity_types/AION.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+
+AION_BS_SS:
+  description: "Basic air ionizer with start/stop"
+  is_canonical: true
+  implements:
+  - AION
+  - SS
+
+AION_BS_SS_SRC_TA_FA:
```
tasodorff, line 27: should be AION_SS_SRC if that is all that it is; the other abbreviations aren't defined as part of the abstract types.

ontology/yaml/resources/HVAC/entity_types/CMP.yaml
```diff
@@ -38,3 +38,13 @@ CMP_SS_WDPM:
   - run_mode
   - schedule_run_command
   - control_status
+
+CMP_SS_ZTO:
```
tasodorff, line 42: shouldnt include an acronym that isnt a defined abstract type. ZTO is undefined.

ontology/yaml/resources/HVAC/entity_types/CMP.yaml
```diff
@@ -38,3 +38,13 @@ CMP_SS_WDPM:
...
+  description: "Basic features of air compressor with zone temp override."
+  is_canonical: true
+  uses:
+  - zone_override_temperature_setpoint
```
tasodorff, line 46: What does this override temperature setpoint mean, and where is the associated setpoint? I am unclear how a compressor does any sort of zone temperature control, which is what this field implies. Please clarify.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -571,4 +571,19 @@ FAN_SS_AL:
   - control_mode
   - run_mode
   - schedule_run_command
-  - control_status
\ No newline at end of file
+  - control_status
+
+FAN_SS_SC_FA_SR:
```
tasodorff, line 576: is this not just a `FAN_SS_VSC`? the field `exhaust_fan_speed_percentage_sensor` should simply be `speed_percentage_command`, since it is obvious from context that it is for an exhaust fan, if the entity name is `EF-<number>`.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -571,4 +571,19 @@ FAN_SS_AL:
...
+  description: "Basic Exhaust fan with start/stop, speed sensor, fail alarm, scheduled run"
+  uses:
+  - exhaust_fan_speed_percentage_sensor
+  implements:
+  - FAN_SS_AL
```
tasodorff, line 581: In general, inheritance from a non-canonical type should be avoided.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -571,4 +571,19 @@ FAN_SS_AL:
...
+  - SRC
+
+FAN_ST:
+  description: "Basic fan with flowrate status only"
+  uses:
+  - supply_air_flowrate_status
```
tasodorff, line 587: If this is directly into a space, it should be `discharge_air_flowrate_status`. It would only be `supply` if it goes to another piece of equipment.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1249,6 +1249,24 @@ FCU_CHWZTC2X_HT2XZTC2X_DFVSC2X:
   - CHWZTC2X
   - HT2XZTC2X
   - DFVSC2X
+  
+FCU_DFVSMC_CHWBRC_STM_RTM_FDPM_ZTC_SRC:
+  description: "FCU with Status, zone control, schedule run"
+  is_canonical: false
+  implements:
```
tasodorff, line 1256: please also add `DFSS` and rename the type accordingly.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1249,6 +1249,24 @@ FCU_CHWZTC2X_HT2XZTC2X_DFVSC2X:
...
+  - FCU
+  - DFVSMC
+  - CHWBRC
+  - STM
```
tasodorff, line 1260: You have DFVSMC called out, so this is a discharge fan -- not a supply fan.  This should be `DTM`.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1249,6 +1249,24 @@ FCU_CHWZTC2X_HT2XZTC2X_DFVSC2X:
   - CHWZTC2X
   - HT2XZTC2X
   - DFVSC2X
+  
+FCU_DFVSMC_CHWBRC_STM_RTM_FDPM_ZTC_SRC:
+  description: "FCU with Status, zone control, schedule run"
+  is_canonical: false
```
tasodorff, line 1255: Why could this not be canonical?

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1249,6 +1249,24 @@ FCU_CHWZTC2X_HT2XZTC2X_DFVSC2X:
...
+  implements:
+  - FCU
+  - DFVSMC
+  - CHWBRC
+  - STM
+  - RTM
+  - FDPM
+  - ZTC
```
tasodorff, line 1263: How does it do zone temperature control if it also does return air temperature control? Which temperature does the valve control to?

ontology/yaml/resources/HVAC/entity_types/HWS.yaml
```diff
@@ -266,4 +266,11 @@ HWS_SWTC_CPC:
...
-  - failed_alarm
\ No newline at end of file
+  - failed_alarm
+  
+HWS_ST:
+  description: "Simple hot water system with status only"
+  implements:
+  - HWS
+  uses:
+  - water_heating_status
```
tasodorff, line 276: `run_command`?

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: I am still unclear what this device does and what data it will provide. It seems like it is a variable-position air damper, with a flow setpoint that can vary between the low- and high-limits. It controls CO2 levels by varying the flowrate setpoint between those limits.

If this is the case, I would expect these fields:
- supply_air_damper_percentage_command
- supply_air_flowrate_sensor
- supply_air_flowrate_setpoint
- low_limit_supply_air_flowrate_setpoint
- high_limit_supply_air_flowrate_setpoint
- zone_air_co2_concentration_sensor
- (plus the schedule fields for SRC)

If this is the case, i would expect it to look like this:
```
implements:
- VAV
- SD
- CO2C
- SRC
uses:
- low_limit_supply_air_flowrate_setpoint
- high_limit_supply_air_flowrate_setpoint
```

You could also create a new abstract type to handle the bounding of the supply_air_flowrate_setpoint between the boundaries like this:
```
HLFC:
  description: "High and low limit bounded supply air flowrate setpoint control."
  uses:
  - low_limit_supply_air_flowrate_setpoint
  - high_limit_supply_air_flowrate_setpoint
  - supply_air_flowrate_setpoint
```

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1079,3 +1083,42 @@ literals:
...
+  - ACTIVE
+  - INACTIVE
+  
+# added for water heating status
+- water_heating_status:
```
tasodorff, line 1116: see comment above about using `run_command` instead. this can be removed.

tasodorff (changes requested):
Another round of clarifying questions and additional comments.

ontology/yaml/resources/HVAC/entity_types/CMP.yaml
josephedwardchang: Hi sir, We use this as ACU (CMP) front to several FCUs

ACU_FCU

I'm asking my engineer colleague if there is specific FCU that this ACU setpoint controls. We may not use it if it's unclear (moving it to opt_uses) but as per specs, this is what it has as its points list (SS, on/off status and this zone override temp setpoint).

ontology/yaml/resources/HVAC/entity_types/AION.yaml
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+########################
+### Canonical Types  ###
+########################
+
+
+AION_BS_SS:
```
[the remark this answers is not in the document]
josephedwardchang: ok sir

ontology/yaml/resources/HVAC/entity_types/AION.yaml
```diff
@@ -0,0 +1,38 @@
...
+  description: "Basic air ionizer with start/stop"
+  is_canonical: true
+  implements:
+  - AION
+  - SS
+
+AION_BS_SS_TA_SR_FA:
```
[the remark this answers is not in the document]
josephedwardchang: noted sir

ontology/yaml/resources/HVAC/entity_types/AION.yaml
josephedwardchang: ok sir. will do as you say.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: FAN_SS_AL is nearer for me sir. will do as you note for speed_percentage_command. Thanks sir

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: noted sir.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: noted sir. Thanks

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: Ah, ok sir, in my case, the points list for this FCU -- we have both return air and supply air so I think I'll use both STM and RTM, for discharge I think it is the zone so ZTC (sensor + setpoint). I'll readjust this to match more closely what we have thanks for the ideas sir.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1249,6 +1249,24 @@ FCU_CHWZTC2X_HT2XZTC2X_DFVSC2X:
...
+  
+FCU_DFVSMC_CHWBRC_STM_RTM_FDPM_ZTC_SRC:
+  description: "FCU with Status, zone control, schedule run"
+  is_canonical: false
+  implements:
+  - FCU
+  - DFVSMC
+  - CHWBRC
+  - STM
+  - RTM
```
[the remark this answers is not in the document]
josephedwardchang: noted sir. Thanks

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: Hi sir, I'm reviewing again the points list and it turns out I need RTM + STM (device has return and supply temp sensors, no control) and then for discharge it is ZTC (zone temp sensor and setpoint).

ontology/yaml/resources/HVAC/entity_types/CMP.yaml
josephedwardchang: I see. Thanks sir. noted on this and will change it to non-canonical

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: no reason sir, was just copied from the other before it hehe. I'll change it.

ontology/yaml/resources/HVAC/entity_types/HWS.yaml
josephedwardchang: in our device points list, there is no run_command sir, just the status if heating or not (no control).

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: ah i see. I'll use run_status sir :) thanks

ontology/yaml/resources/HVAC/entity_types/CMP.yaml
josephedwardchang: Hi again sir,

I was wrong, document we have as points list has typo: ACU has only SS (not including zone temp override setpoint). the zone temp override setpoint got a corrputed header it is part of the occupancy sensor (which has been removed now) and so the point list under occupancy sensor (in excel) got into the group above it when occupancy header got removed. I'll revert this ASAP.

Thanks for catching it sir, I am not really accustomed to hardwares like this, as I am a software developer by trade hehe.

ontology/yaml/resources/HVAC/entity_types/CMP.yaml
josephedwardchang: removing this now as we will be using CMP_SS only

ontology/yaml/resources/HVAC/entity_types/ADY.yaml
josephedwardchang: noted sir.  I changed some more items to reflect this. thanks.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: Thanks sir, the final one is much more simplified now:

```
VAV_SDM_CO2C_SRC
  description: "CO2 controlled VAV, with damper, hi/low flowrate sensor and setpoint and schedule run"
  is_canonical: true
  implements:
  - VAV
  - SDM
  - CO2C
  - SRC
  uses:
  - low_limit_supply_air_flowrate_setpoint
  - high_limit_supply_air_flowrate_setpoint
  - supply_air_flowrate_sensor 
```
Thanks

ontology/yaml/resources/HVAC/entity_types/HWS.yaml
```diff
@@ -325,4 +325,11 @@ HWS_SWTC_CPC:
   uses:
   - circulation_pump_speed_percentage_command
   - circulation_pump_speed_percentage_setpoint
-  - failed_alarm
\ No newline at end of file
+  - failed_alarm
+  
+HWS_ST:
```
tasodorff, line 330: please make this HWS_SS, implement SS below, and remove the field usage. in fact, the type `HWS_SS_ALS` already exists so I suggest using that.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: I am still confused on control logic. You have used `SDM`, which is two-position control. How does it control to a flow setpoint with two-position control on the damper? And how do the low and high limits relate to the target setpoint, and where is the target setpoint?

ontology/yaml/resources/METERS/entity_types/ABSTRACT.yaml
```diff
@@ -20,8 +20,8 @@ PWM:
   description: "Very basic power and enrgy monitoring only."
   is_abstract: true
   opt_uses:
-  - apparent_power_sensor
-  - reactive_power_sensor
+  - apparentpower_sensor
```
tasodorff, line 23: Please revert. [This PR](https://github.com/google/digitalbuildings/pull/581) will be removing `apparentpower` and `reactivepower` from the ontology.

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
```diff
@@ -19,4 +19,15 @@ FD_STD:
...
-  - damper_status
\ No newline at end of file
+  - damper_status
+  
+ FD_SD:
+  description: "Basic smoke detector fire damper"
+  is_canonical: true
+  implements:
+  - FD
+  - SD
```
tasodorff, line 29: You cannot implement this because it is not in the same namespace. You also have defined damper controls below, but omitted the `supply`. You intend to have a device with both `supply_air_damper_percentage_command` and `damper_percentage_command`? Please explain.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -471,7 +471,7 @@ literals:
 - zone_air_refrigerant_concentration_setpoint
 
 # Added on 2019/06/03
-- apparent_power_sensor
+- apparentpower_sensor
```
tasodorff, line 474: revert

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -492,7 +492,7 @@ literals:
 - phase3_power_sensor
 - phase3_powerfactor_sensor
 - powerfactor_sensor
-- reactive_power_sensor
+- reactivepower_sensor
```
tasodorff, line 495: revert

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1086,6 +1090,41 @@ literals:
...
+  - INACTIVE
+- water_flowrate_status:
+  - ACTIVE
+  - INACTIVE
+- air_pressure_status:
+  - ACTIVE
+  - INACTIVE
+  
+# added for ACU type
+- zone_override_temperature_setpoint
```
tasodorff, line 1123: remove, this is not referenced elsewhere in the PR.

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
```diff
@@ -138,3 +138,25 @@ LC:
...
   - illuminance_sensor
+
+DFRT:
+  description: "Lighting Control dimmer fade rate & time"
+  is_abstract: true
+  uses:
+  - dimmer_fade_rate_setpoint
+  - dimmer_fade_time_setpoint
+
+ZLC:
```
tasodorff, line 149: I take issue with most of these names. Isnt `lighting` redundant since it is associated with a lighting system? Secondly, what devices do you see this being applied to? Gateways? Luminaires?

Additionally, what is a lighting switch? A wall switch? If so, why does run_command not work?

Please explain.

tasodorff (changes requested):
Please fix validator errors and address comments.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -614,6 +614,38 @@ FAN_SS_AL:
   - schedule_run_command
   - control_status
 
+FAN_SS_VSC_SRC:
+  description: "Basic Exhaust fan with start/stop, speed sensor, fail alarm, scheduled run"
+  uses:
+  - lost_power_alarm
+  opt_uses:
+  - filter_alarm
```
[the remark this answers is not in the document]
josephedwardchang: Hi sir, we have another FAN device that has the same points and an additional filter_alarm, so rather than create another canonical type with filter_alarm, I decided they should share the same type and both can use it.

ontology/yaml/resources/HVAC/entity_types/HWS.yaml
josephedwardchang: I see. Will do. Thanks.

ontology/yaml/resources/METERS/entity_types/ABSTRACT.yaml
josephedwardchang: Noted sir. Thanks

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: Noted sir. Thanks

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: Noted sir. Thanks

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: sorry yes, this will be for upcoming pull request, together with some others but I am still working on them sir.

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
josephedwardchang: ah yes, totally forgot about this. this was from older pull request #386, but we will not be including Ligthing data for now. I am removing all Lighting changes.  

I was still learning DBO when I had this (oct '21) and my first impression was to describe the field accurately, i.e., zone_lighting_switch_command for describing turning on/off lights in an area (zone). It was fairly recent that I learned about run_command/status for the same purpose. You would not believe, I had previously used "power_command" (or something to that effect) until you pointed out power is not the correct one for this hehe.  

much thanks sir.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: I see what you mean. I'll ask again from my colleagues.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: Not sure about chilled water sir. It is Pre-Cooled Fan Coil Units (Ceiling Concealed Ducted Type) but here is where it is confusing me too: it has 2 point list for zones that is described as "Zone Air Temperature", "Zone Temperature Set Point Adjustment/Control". the 1st one is sensor, but the 2nd one is a combined setpoint and command. so, CHWZTC for me is not an option(?) as it has 3 points for zone. I'm not sure what to do here.

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
josephedwardchang: Oh hehe. I saw in SAFETY\GENERALTYPES the abstract type "SD" as smoke detector, which is what I needed sir :) this is correct, i think, fire damper that's a smoke detector, it is nearest for the device we have.

tasodorff (commented):
See last comment -- good to go once clarified.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -614,6 +614,38 @@ FAN_SS_AL:
...
+  implements:
+  - FAN
+  - SS
+  - VSC
+  - SRC
+  - DPM
+FAN_ST:
+  description: "Basic fan with flowrate status only"
+  uses:
+  - air_flowrate_status
```
tasodorff, line 645: This field is undefined and generating validator [output errors](https://github.com/google/digitalbuildings/runs/6578470018?check_suite_focus=true). Please fix.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: You are correct sir on SDM. Will this be ok?

```
VAV_CO2C_SRC:
  description: "CO2 controlled VAV, with damper, hi/low flowrate sensor and setpoint and schedule run"
  is_canonical: true
  implements:
  - VAV
  - CO2C
  - SRC
  uses:
  - low_limit_supply_air_flowrate_setpoint
  - high_limit_supply_air_flowrate_setpoint
  - supply_air_flowrate_sensor 
  - supply_air_damper_command
  - supply_air_damper_percentage_sensor
```

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: Not sure about chilled water sir. It is Pre-Cooled Fan Coil Units (Ceiling Concealed Ducted Type) but here is where it is confusing me too: it has 2 point list for zones that is described as "Zone Air Temperature", "Zone Temperature Set Point Adjustment/Control". the 1st one is sensor, but the 2nd one is a combined setpoint and command. so, CHWZTC for me is not an option(?) as it has 3 points for zone. I'm not sure what to do here.

this is the full point list for Pre-Cooled Fan Coil Units (Ceiling Concealed Ducted Type)

```
Supply Air Temperature
Return Air Temperature
Filter Status
Unit Enable
Unit Status
Unit Fault
Zone Air Temperature
Zone Temperature Set Point Adjustment/Control
Time Schedule
```

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
josephedwardchang: yes sir it is contained within but no data other than dampers are available


Motorized Smoke Fire Damper
```
Open/Close Feedback
Open/Close Control
```

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: fixed sir.

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
tasodorff: im ok with FD STD. we can rename later if necessary.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -482,15 +482,16 @@ VAV_RD_CO2C:
...
   implements:
   - VAV
-  - SDM
   - CO2C
   - SRC
   uses:
   - low_limit_supply_air_flowrate_setpoint
   - high_limit_supply_air_flowrate_setpoint
   - supply_air_flowrate_sensor 
+  - supply_air_damper_command
```
tasodorff, line 496: Why does this have a two position damper and a percentage sensor? Is that intended?

Also, where is the supply air flowrate setpoint? I see the high/low limits, but I do not see the actual setpoint. 

If this is intended to model new functionality (i.e. there is no existing abstract type that covers this) we should create it; if its unique and we don't wish to give it an abstract type, it should be non-canonical (`is_canonical: false`). But please advise on the questions above first, because I do not understand the device's operation..

tasodorff (commented):
Please see comment on most recent VAV addition.

tasodorff:
Please also fix [errors](https://github.com/google/digitalbuildings/runs/6594802519?check_suite_focus=true).

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: These are the points list I have sir. I contacted our engineers for the device details/operation (dampers) and they have yet to get back to me on this. Can I ask your help and infer from this list? the only setpoint I see is the CO2, is it possible the CO2 has relationship with the dampers? 

```
VAV
--
CO2 Setpoint
Minimum Air Flow Rate Setpoint
Maximum Air Flow Rate Setpoint
Air Flow Rate Present Value
VAV Damper Control
VAV Damper Feedback
Time Schedule
```

Thanks sir

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -614,6 +614,38 @@ FAN_SS_AL:
...
+  - lost_power_alarm
+  opt_uses:
+  - filter_alarm
+  implements:
+  - FAN
+  - SS
+  - VSC
+  - SRC
+  - DPM
+FAN_ST:
```
tasodorff, line 642: please add blank line above.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
tasodorff: It seems like your fieldset is incomplete. The abstract types I specified above seem to be correct, but this particular device doesn't provide you with chilled water valve control directly (it does do chilled water cooling, you just don't have points for the valve). This is an example of an incomplete translation. In your building config, you would specify that it is this type (including CHWZTC) but you would mark those specific fields as `MISSING`. Does that make sense? You can see an example of it [here](https://github.com/google/digitalbuildings/blob/6e597b34556d71d9c9c1f1f835819526601b5da0/ontology/docs/building_config.md#defining-translations).

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
tasodorff: you need to use discharge air temperature though, because it is discharging into a space -- not being supplied to other equipment. its a subtle difference, but it is important for us. please use DTM and RTM.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
tasodorff: same comment as above. it sounds like this is CHW cooling but you dont have the data for the chilled water valve. If its acually DX then you would use the DX equivlent (DXZTC), but would still have missing compressor information. This is totally fine, it just needs to reflect what the equipment actually is, even if telemetry is missing.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: in this case, wouldnt it simplify even further to `VAV_SD_HLFC_CO2C_SRC`? you also still need the flowrate setpoint, which i assume you have because it is bounded by the high and low limits.

again, perfectly acceptable to have `MISSING` fields in your translation if the data is not available.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: I see sir. Ok, I can do that. will adjust now to what you recommend, thanks sir

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: will also adjust now to what you recommend, thanks sir

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: understood sir. I see your point about MISSING and I did not look into SD and HLFC further hehe. Thanks again.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: yes sir, didn't notice it sorry

mschulze17:
@tasodorff back in your court

ontology/yaml/resources/METERS/entity_types/EM.yaml
```diff
@@ -28,6 +28,17 @@ EM_PWM_PVCM:
   - EM
   - PWM
   - PVCM
+  opt_uses:
+  - reactive_energy_accumulator
+  - apparent_energy_accumulator
+
+EM_PWM_AVCM:
```
tasodorff, line 35: Cannot have a duplicate, see line 43.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
tasodorff: please resolve this comment when you have made the changes

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: @josephedwardchang what is the action here? i still expect that this be adjusted based on the comments above.

ontology/yaml/resources/SAFETY/entity_types/LDS.yaml
```diff
@@ -85,3 +85,11 @@ LDS_WLA_BCM:
   opt_uses:
   - water_leak_cable_distance_sensor
 
+SWV_CM:
```
tasodorff, line 88: If this is not an LDS then it doesnt belong in the LDS file. Please move to SWV.yaml.

tasodorff (changes requested):
Still requires changes and updates, and must pass validation. See comments.

ontology/yaml/resources/METERS/entity_types/EM.yaml
josephedwardchang: done sir, although i don't know how this could happen. probably through the auto merge...

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: yes sir. that's what I came up with. the HLSAFS contains the needed points. and I added 2 sensors for waht we need: co2 and flowrate sensors.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
josephedwardchang: will do sir. this is what we came up with. we can't use CHW and these below are the closest to waht we need, with additional points that will be marked MISSING. thanks sir
```
FCU_RTM_DXZTC_SRC:
  description: "FCU with SS, supply and return temp sensors, zone control and sensor, schedule run, filter change and general fail alarm"
  is_canonical: true
  implements:
  - FCU
  - RTM
  - DXZTC
  - SRC
  uses:
  - filter_alarm
  - lost_power_alarm
```

ontology/yaml/resources/SAFETY/entity_types/LDS.yaml
josephedwardchang: sir, I found VLV for valves is in HVAC, so I moved this to VLV.yaml and will now define SWCM for water solenoid valves in abstract. ultimately will use VLV_SWCM for sommand and status monitoring. Thanks

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -4966,4 +4975,13 @@ EFHLC:
   - low_exhaust_fan_speed_status
   - high_exhaust_fan_speed_status
   implements:
-  - OPERATIONAL
\ No newline at end of file
+  - OPERATIONAL
+
+SWCM:
```
tasodorff, line 4980: i would be expecting valve information on the fields, not run commands (which are used to tell pieces of equipment to operate). also, without some context for the devices themselves its hard to say what the least necessary set of subfields will be. is this a stand-alone solenoid valve (and is it really in the HVAC namespace) or is it part of another piece of equipment.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: i am still confused on the function here. i don't understand how you can have a two-position damper with flowrate setpoints... it won't modulate to control airflow to a specific value, it will just be whatever value it is based on geometry of the box and back pressure in the duct. does the box use two-position damper control (as indicated by the description) or does it use modulating, continuous control (that is, will it have `supply_air_damper_command` or `supply_air_damper_percentage_command`)?

 also, for co2, why not implement CO2C?

ontology/yaml/resources/HVAC/entity_types/VLV.yaml
```diff
@@ -58,4 +58,9 @@ VLV_CDWPVM:
...
   - CDWPVM
 
-  
\ No newline at end of file
+VLV_SWCM:
+  description: "Solenoid water valve control and monitoring."
+  is_canonical: true
+  implements:
+  - VLV
+  - SWCM
```
tasodorff, line 66: what kind of valve is it? what system does it live on? perhaps one of the existing types covers this better? i am concerned with using terms like `run_command` and `run_status` here. If we need more generic terms for the valve command, i suggest `valve_command` and `valve_status`, even though it deviates slightly from our other conventions.

tasodorff (commented):
please see comments for clarification.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: Hi again sir,

This is the list of points for the VAV we need:
```
CO2 Setpoint		
Minimum Air Flow Rate Setpoint		
Maximum Air Flow Rate Setpoint		
Air Flow Rate Present Value		
VAV Damper Control	Modulating Damper	%
VAV Damper Feedback	Modulating Damper	%
Time Schedule
```
I may have been wrong on the dampers as they are clearly % (sorry for that), does the list seem valid? I have no idea if it is, as CO2 is stated as setpoint but doesn't contain CO2 sensor. With your last comments I'm leaning towards this now:

```
VAV_PDSCV_CO2C_HLSAFS_SRC:
  description: "VAV with CO2 monitor, dampers, hi/low flowrate setpoint and sensor, schedule run"
  is_canonical: true
  implements:
  - VAV
  - PDSCV
  - CO2C
  - HLSAFS
  - SRC
  uses:
  - supply_air_flowrate_sensor
```
Thanks

ontology/yaml/resources/HVAC/entity_types/VLV.yaml
josephedwardchang: Hi sir,
the valve (solenoid water valve) is independent and will be controlled through BMS. I'll use yuor suggestions, thanks.

tasodorff:
@josephedwardchang please fix validator errors.

josephedwardchang:
> @josephedwardchang please fix validator errors.

updated to fix errors sir. Thanks

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -664,6 +664,36 @@ FAN_SS_AL:
...
+  - VSC
+  - SRC
+
+FAN_VSC_SRC_DPM:
+  description: "Basic Exhaust fan with start/stop, speed sensor, fail alarm, scheduled run, dampers"
+  uses:
+  - lost_power_alarm
+  opt_uses:
+  - filter_alarm
+  implements:
```
tasodorff, line 684: should also include `SS`

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -664,6 +664,36 @@ FAN_SS_AL:
...
+  - FAN
+  - VSC
+  - SRC
+  - DPM
+
+FAN_PS:
```
tasodorff, line 690: this fan doesnt have anything other than an air pressure sensor? we should model this as another type of fan like FAN_SS and add air pressure sensor as optional.

ontology/yaml/resources/METERS/entity_types/WM.yaml
```diff
@@ -26,9 +26,10 @@ WM_STANDARD:
...
-  description: "Water supply flow rate meter."
+  description: "Water supply flow rate meter with counter."
   implements:
   - WM
   is_canonical: true
   uses:
-  - supply_water_flowrate_sensor
\ No newline at end of file
+  - supply_water_flowrate_sensor
+  - water_volume_accumulator
```
tasodorff, line 35: supply_water_volume_accumulator

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: it is missing its supply_air_flowrate_setpoint? i cannot have setpoint boundaries without the septoint itself. i also think it needs to use SD instead of PDSCV.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -511,6 +511,18 @@ VAV_RD_CO2C:
   - RD
   - CO2C
 
+VAV_PDSCV_CO2C_HLSAFS_SRC:
```
tasodorff, line 514: given the other comment string, i think PDSCV gets replaced with SD and we are good here. use MISSING translation values to deal with data that it does not send us.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -831,7 +834,23 @@ literals:
...
+- valve_command:
+  - OPEN
+  - CLOSED
+- valve_status:
+  - OPEN
+  - CLOSED
+- low_battery_alarm:
+  - ACTIVE
+  - INACTIVE
+- load_bypass_mode:
```
tasodorff, line 847: i did not see this used elsewhere in the PR. can you please tell me where it is used? any unused fields should be omitted.

tasodorff (changes requested):

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: Hi sir, I found VSC already has SS (run_command and run_status)

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: Noted sir. will do.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: you're right sir, I didn't see this combination before hehe. Thanks, will do this.

ontology/yaml/resources/METERS/entity_types/WM.yaml
josephedwardchang: i thought I should just use existing, but i see now. Thanks sir, will do.

ontology/yaml/resources/fields/telemetry_fields.yaml
josephedwardchang: oops, this will be for next pull req, sorry. removed.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
josephedwardchang: you're right sir, I didn't see this combination before hehe. Thanks, will do this.

josephedwardchang:
I need a little help sir, my validation has one error:

```
Starting config validation...

Validating entities ...
Field /apparent_power_sensor_6 has an undefined measurement unit: kilovolt_amperes_apparent
Field /reactive_power_sensor_13 has an undefined measurement unit: kilovolt_amperes_reactive
```

but that unit is in "apparentpower", (now, i know it is to be reverted) and if I use it, my error would then be "apparentpower_sensor_6" is missing (or some such same meaning as undefined, nothing, etc). Same goes for "reactive_power_6". What do I need to do here? this is under Energy Meters: EM_PWM_PVCM

josephedwardchang:
Hi again sir, is there a way to create a new pull request while I have a pending one? I believe my CLA is used with my email for the github forked codes that I am not able to crete a new one. Or is the pull request by design to only have one pull request per fork?

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
tasodorff: understood but we usually include both. if you wouldnt mind adding it, it will make it a bit more consistent with the others.

ontology/yaml/resources/METERS/entity_types/WM.yaml
```diff
@@ -26,9 +26,10 @@ WM_STANDARD:
 
 WM_WF:
```
tasodorff, line 27: this will invalidate the existing type if we add a new required field (existing meters of this type will all be broken). please add the accumulator as optional, or create a new type with both required.

tasodorff (changes requested):
one change to the fan and i think we are good to go. thank you for struggling through this with me!

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
josephedwardchang: Oh I tested with DBO telemetry validation and including SS did not have error. I was expecting it errors out something like "run_command is already defined", but no it didn't. I am adding SS now sir.

ontology/yaml/resources/METERS/entity_types/WM.yaml
josephedwardchang: got it. thanks sir.

tasodorff (approved):

mschulze17:
@cstirdivant yours to merge

PR #488, opened 2022-02-08
Addition of liters_per_hour unit

ghairapetian:
Addition of a unit for a Heat meter


tasodorff (approved):

ghairapetian:
Hi, every time it's approved, I need to update and therefore, it needs approval again. How can we manage that ?

PR #489, opened 2022-02-09
Instance Validator Directory Reference

tasodorff:
Updated the instance validator to also accept directories. You can now pass any combination of (mutually exclusive) directories and files to the validator and it will read them all into a single instance validation run.

Also made a few minor fixes elsewhere:
- entity_instance.py -- fixed error warning not printing the bad units properly
- telemetry.py -- default to version 1 if no version is provided in the payload


tools/validators/instance_validator/validate/handler.py
```diff
@@ -108,7 +118,32 @@ def RunValidation(filenames: List[str],
...
+      else:
+        unpacked_files.append(file)
+
+    # Because we are recursively walking the tree, we need to check
+    # that they didnt accidentally pass in a directory that contains
+    # a subdirectory they also passed. This will lead to stupid and
+    # unhelpful key errors if left unchecked. Best to do that here
+    # and save the troubleshooting headache. We could just deduplicate
+    # but this is probably the safer route to go.
+    for elem in unpacked_files:
```
dplute, line 138: An alternative to consider is to change unpacked_files to a set instead of manually checking list elements. The _ValidateConfig and Deserialize type hints would need to be changed to Iterable instead of List, but that's fine since we only use the collection of filenames via iteration. It wouldn't be able to report that a folder was included multiple times, but it also wouldn't encounter issues if that happened.

tools/validators/instance_validator/validate/handler.py
```diff
@@ -79,6 +80,15 @@ def _ValidateTelemetry(subscription: str, service_account: str,
   helper = TelemetryHelper(subscription, service_account)
   helper.Validate(entities, timeout)
 
+def _GetFilepathsFromDir(root_dir):
+  """ Takes in a directory and returns a filepath list for all YAML
+  files it finds. """
+  file_paths = []
+  for root, _, files in os.walk(root_dir, topdown=False):
+    for name in files:
+      if '.yaml' in name:
```
dplute, line 89: Minor, but `name.endswith('.yaml')` is more specific to what we're checking for.

dplute (approved):

charbull:
can you please undo the deletion of this file: .gitignore

tools/validators/instance_validator/tests/fake_resources/fakedir/another_fan/ef2.yaml
```diff
@@ -0,0 +1,19 @@
+EF - 2 Restroom / 1:
+  cloud_device_id: '3'
+  guid: 12345
+  connections:
+    US-MTV-1:
+    - CONTAINS
+  id: CDM/3
```
charbull, line 7: this is no longer supported, can you please remove

tools/validators/instance_validator/tests/fake_resources/fakedir/bldg.yaml
```diff
@@ -0,0 +1,4 @@
+US-MTV-1:
+  id: FACILITIES/buildings/1
```
charbull, line 2: same here

tools/validators/instance_validator/tests/fake_resources/fakedir/fans/ef3.yaml
```diff
@@ -0,0 +1,19 @@
+EF - 3 Restroom / 1:
+  cloud_device_id: '2'
+  guid: 12345963
+  connections:
+    US-MTV-1:
+    - CONTAINS
+  id: CDM/2
```
charbull, line 7: ditto

charbull (changes requested):
just nits for the test files.

tasodorff:
@charbull changes made

charbull (approved):

PR #515, opened 2022-03-24
Update subfields.yaml

gerrykou:
Added ```condensation``` as a ```measurement_descriptor``` e.g ```condensation_sensor```


tasodorff (commented):
Does this solve your use case:
- [condensing](https://github.com/google/digitalbuildings/blob/7c1c952dc9a253ee5abf190bb98c097c89757471/ontology/yaml/resources/subfields/subfields.yaml#L78)

(we can adjust the definition to include more than just refrigerant.)

noursaidi:
I think the project specific use case behind this PR was around water condensing on the from chilled beams, where so called _'condensation sensors_' are used to as a switch to turn off the a specific chilled beam or zone when moisture is detected or a risk of condensation forming. These may be a a moisture (condensate) detector on the pipe, a dew point switch on the face of the chilled beam, or a calculation from the zone T+%RH sensor.

I'm looking to create a PR to add chilled beams as an entity type - may be best to address this query in that

tasodorff:
> 

Perhaps it would be. I need to see some additional context to determine where the the new subfield would be necessary, as distinct from `condensing`. We can always adjust the definition of `condensing` to make it more general (i.e. remove the reference to refrigerant gas and apply it to all condensing gas).

tasodorff (approved):

mschulze17:
Assigning to @cstirdivant for merge

PR #519, opened 2022-03-30
Entity connection identification test; define dimension category on classes

ahemphill:


tools/scoring/tests/dimensions/standard_field_naming_test.py
```diff
@@ -17,7 +17,7 @@
 from absl.testing import absltest
 
 from score.dimensions.standard_field_naming import StandardFieldNaming
-from score.constants import FileTypes
+from score.constants import FileTypes, DimensionCategories
```
charbull, line 20: can you add the constants to the PR? or I am not looking at the right spot?

charbull (changes requested):

tools/scoring/tests/dimensions/standard_field_naming_test.py
ahemphill: They're at `tools/scoring/score/constants.py`

tools/scoring/score/dimensions/entity_connection_identification.py
```diff
@@ -25,6 +25,9 @@
 class EntityConnectionIdentification(Dimension):
   """Quantifies whether connections between entities were
   correctly and completely defined in the proposed file."""
+
+  category = DimensionCategories.COMPLEX
```
charbull, line 29: how do you determine if it is a simple or complex one for a connection?

tools/scoring/score/dimensions/entity_identification.py
```diff
@@ -27,6 +27,9 @@
 class EntityIdentification(Dimension):
   """Quantifies whether the correct entities
   were included in the proposed file."""
+
+  category = DimensionCategories.COMPLEX
```
charbull, line 31: same here how do you know it is Complex?
do you need to add extra explanation here? 
https://github.com/google/digitalbuildings/blob/master/tools/scoring/score/constants.py#L26

tools/scoring/score/dimensions/entity_point_identification.py
```diff
@@ -26,6 +26,9 @@
 class EntityPointIdentification(Dimension):
   """Quantifies whether the proposed file
   included the correct points in each entity."""
+
+  category = DimensionCategories.COMPLEX
```
charbull, line 30: only translations are simple?
do we need to distinguish?

tools/scoring/score/dimensions/unit_mapping.py
```diff
@@ -14,14 +14,17 @@
...
+from score.constants import FileTypes, DimensionCategories
 
 PROPOSED, SOLUTION = FileTypes
 
 
 class UnitMapping(Dimension):
   """Quantifies how accurately the proposed file
   mapped dimensional units for relevant fields."""
+
+  category = DimensionCategories.SIMPLE
```
charbull, line 26: can you please update https://github.com/google/digitalbuildings/blob/master/tools/scoring/score/constants.py#L26
to clarify which ones are simple and which are complex?

It seems you differentiate them based on how you are calculating them?

If it is simple ---> simple processing ?
complex --> the aglorithm needs to do some extra steps to identify things?

tools/scoring/score/parse_config.py
```diff
@@ -198,43 +198,35 @@ def retrieve_reporting_translations(
...
-    for dimension_category, dimension_list in dimensions.items():
+    for dimension in dimensions:
       # Invoke the functions and append the dictionary with their return values
-      for dimension in dimension_list:
-        if dimension_category == SIMPLE:
-          invoked = dimension(translations=translations).evaluate()
-        elif dimension_category == COMPLEX:
-          invoked = dimension(deserialized_files=deserialized_files).evaluate()
+      if dimension.category == SIMPLE:
+        invoked = dimension(translations=translations).evaluate()
```
charbull, line 225: nit: `evaluated` ?

tools/scoring/tests/dimensions/entity_connection_identification_test.py
```diff
@@ -12,41 +12,187 @@
...
+        proposed_path=featureful_file_path,
+        solution_path=featureful_file_path)
+
+    empty_file_path = 'tests/samples/empty.yaml'
+    self.none_score_argument = TestHelper.prepare_dimension_argument(
+        dimension=EntityConnectionIdentification,
+        proposed_path=empty_file_path,
+        solution_path=empty_file_path)
+
+    no_contains_file_path = 'tests/samples/proposed/entity_connection_identification_no_contains.yaml'  # pylint: disable=line-too-long
```
charbull, line 43: why do we care of the specific case `no_contains` ?

tools/scoring/tests/dimensions/entity_identification_test.py
```diff
@@ -31,23 +30,26 @@ def setUp(self):
     featureful_file_path = (
         'tests/samples/proposed/entity_identification_virtual_guid.yaml')
     self.highest_score_argument = TestHelper.prepare_dimension_argument(
-        dimension_type=COMPLEX,
+        dimension=EntityIdentification,
```
charbull, line 33: this is a bit confusing?
dimension is either simple or complex?

charbull (changes requested):

tools/scoring/score/dimensions/entity_connection_identification.py
ahemphill: Connections are a feature of both reporting and virtual entities, so to score them we need to pass the deserialized files rather than the pre-matched list of translations. Simple dimensions receive translations whereas complex dimensions receive deserialized files.

tools/scoring/score/dimensions/entity_identification.py
ahemphill: Commented above; it is complex because the score is derived from information which is effectively not processed beforehand. More context is provided in `parse_config.py` and in `dimension.py`, and in fact the comment in `constants.py` is not strictly accurate (as a multi-map is not built for Entity Connection Identification). I will work on expanding the explanation and unifying it between files.

tools/scoring/score/dimensions/entity_point_identification.py
ahemphill: Correct. I can comment above each of these assignments.

tools/scoring/score/dimensions/unit_mapping.py
ahemphill: You got it. I'll work on expanding and unifying the explanation.

tools/scoring/tests/dimensions/entity_connection_identification_test.py
ahemphill: It's arbitrary — just what I eliminated to make fewer connections. Any naming suggestions?

tools/scoring/tests/dimensions/entity_identification_test.py
ahemphill: Yes, and now it is defined on the classes to avoid the overhead of maintaining separate lists.

tools/scoring/score/dimensions/entity_connection_identification.py
charbull: cool can you update the comment for the definitions of Simple vs complex with your comment above ? :)

tools/scoring/score/dimensions/entity_identification.py
charbull: thanks !

tools/scoring/tests/dimensions/entity_identification_test.py
charbull: I see, 
since `dimension is either simple or complex by definition` should we give another name for `dimension` in`dimension=EntityIdentification`?

tools/scoring/score/dimensions/entity_identification.py
charbull: are you going to do it in this PR? or have a todo ?

charbull (commented):
questions

tools/scoring/score/dimensions/entity_identification.py
ahemphill: I'll do it here.

tools/scoring/tests/dimensions/entity_identification_test.py
ahemphill: I could change the argument to `dimension_category` but I'm not sure that would add clarity, e.g.: `dimension_category=EntityIdentification.category`. The way it stands, the retrieval of the `category` value is abstracted similarly to how it is in `parse_config.py`.

charbull (approved):

PR #525, opened 2022-04-06
Geothermal Loop Additions

cstirdivant:
New subfields, fields and heat meter types for the geothermal loop at MFA-BV200


tasodorff:
@cstirdivant please review validator errors.

ontology/yaml/resources/METERS/entity_types/HM.yaml
```diff
@@ -48,3 +48,23 @@ HM_HW:
...
+
+HM_HW_WAC:
+  description: "Heating load heat meter with total heating water volume accumulator."
+  implements:
+  - HM_HW
+  is_canonical: true
+  uses:
+  - water_volume_accumulator
+
+HM_GTW:
```
tasodorff, line 60: since this is clear it applies to the geothermal loop (because the meter will have a name like `HM-1_GTWL` or something) is it necessary to include the geothermal qualifier on this device? it is not distinguishing it from anything else on that particular device.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1079,3 +1079,18 @@ literals:
...
+## Geothermal loop fields
+- geothermal_return_water_temperature_sensor
+- geothermal_supply_water_temperature_sensor
+- geothermal_thermal_energy_accumulator
+- geothermal_thermal_power_sensor
+- geothermal_water_settling_temperature_sensor
+- geothermal_water_flowrate_sensor
+
+## PV meter fields
+- photovoltaic_cell_charge_energy_accumulator
```
tasodorff, line 1091: is a use for these fields forthcoming? i would not want to commit additional fields without downstream utilization in types.

tasodorff (commented):
A few comments, please take a look and clarify.

PR #532, opened 2022-04-11
Raw field selection test

ahemphill:


tools/scoring/tests/dimensions/raw_field_selection_test.py
```diff
@@ -19,35 +19,129 @@
...
 
 class RawFieldSelectionTest(absltest.TestCase):
   def setUp(self):
     super().setUp()
-    # TODO: add real data (append cases to existing tests)
-    translations = {'cloud_device_id': {PROPOSED: [], SOLUTION: []}}
-    self.raw_field_selection = RawFieldSelection(
-        translations=translations).evaluate()
+    one_entity_file_path = (
+        'tests/samples/proposed/raw_field_selection_one_entity.yaml')
```
charbull, line 31: can we reuse existing files? they seem the same?

charbull (commented):
question: can we reuse the test files instead of new ones?

tools/scoring/tests/dimensions/raw_field_selection_test.py
ahemphill: They have the same content in this circumstance but I maintain that they should be separate for a few reasons:
- The purpose of unit tests is isolation, and combining them breaks that pattern
- Combining them, especially when the content does not overlap, makes it very difficult to reason about what is being tested and hampers maintainability
- Given that some tests require their own files, discerning the reasoning behind the inconsistency will be challenging down the line

What advantage do you think combining them would provide?

charbull: repeating unnecessary content?

charbull (commented):
these files are the same, why not reuse them?

tools/scoring/tests/dimensions/raw_field_selection_test.py
ahemphill: Updated the files to differentiate them.

charbull (approved):

PR #536, opened 2022-04-12

trav3711:
PiperOrigin-RevId: 441020069

### DIFF
--- a/.github/workflows/python_tools.yml
+++ b/.github/workflows/python_tools.yml
@@ -51,18 +51,6 @@ jobs:
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/explorer
-    #---------- Configuration scoring ----------#
-    - name: Configuration scoring Install dependencies
-      run: |
-        python setup.py install
-      working-directory: ./tools/scoring
-    - name: Run Pylint on Configuration Scorer
-      run: |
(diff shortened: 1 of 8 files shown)


trav3711:
Can we force this through? The things it is failing on are Instance Validator linter errors which weren't thrown by the g3 linter...

charbull:
@trav3711 please fix linter

.github/workflows/python_tools.yml
```diff
@@ -51,6 +51,18 @@ jobs:
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/explorer
+    #---------- Configuration scoring ----------#
```
charbull, line 54: undo? keep the scorer after the validator since there is a dependency

charbull (changes requested):
fix linter please

.github/workflows/python_tools.yml
trav3711: Ah ok, this is incorrect in g3 as well then. Will remove.

charbull: did you remove? I am not sure if I am looking at the latest version?

charbull (changes requested):
revert the github actions file please

charbull (approved):

PR #546, opened 2022-04-20
UX tweaks

ahemphill:


tools/scoring/README.md
```diff
@@ -11,15 +11,15 @@ To install the dependencies, please run the `python3 setup.py install` from the
 ## Usage
 
 The application can be run from the command line. It takes three arguments:
-1. `-ont/--ontology` (required): Absolute path for the directory which contains your ontology
-2. `-prop/--proposed` (required): Absolute path for your proposed configuration file (to be scored)
-3. `-sol/--solution` (required): Absolute path for your solution configuration file
+1. `-prop/--proposed` (required): Absolute path for your proposed configuration file (to be scored)
+2. `-sol/--solution` (required): Absolute path for your solution configuration file
+3. `-ont/--ontology` (optional): Absolute path for the directory which contains your ontology. Defaults to `ontology/yaml/resources`
```
charbull, line 16: can you name it as in the ontology-validator please

tools/scoring/score/parse_config.py
```diff
@@ -82,6 +84,7 @@ def append_types(
         The deserialized files dictionary with types
         appended to each entity in the respective files.
     """
+    print('Scoring — appending entity types')
```
charbull, line 87: I dont think this is needed? it is an implementation detail?

tools/scoring/score/parse_config.py
```diff
@@ -199,6 +202,7 @@ def aggregate_results(
     results = {}
 
     for dimension in dimensions:
+      print(f'Scoring — calculating {dimension.__name__}')
```
charbull, line 205: this might be too much verbose?

charbull (changes requested):

tools/scoring/score/parse_config.py
ahemphill: Fair point.

ahemphill: Well, it contextualizes that follows, which _is_ relevant.

charbull (approved):

PR #568, opened 2022-04-28
Addition of VAV_RDM and RDM function

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -4733,3 +4733,12 @@ HWRWISOVPM:
   - heating_return_water_isolation_valve_percentage_sensor
   implements:
   - MONITORING
+
+RDM:
```
tasodorff, line 4737: RDC would be more appropriate, because you are commanding the return damper. Please rename.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -388,6 +388,13 @@ VAV_DSP_ED:
   - DSP
   - ED
 
+VAV_RDM:
```
tasodorff, line 391: This device does not have flow control? How does it know what to set the damper to? What does it control to? It should have control.

tasodorff (changes requested):

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
ghairapetian: Actually, there are two VAV, one at the supply side that control the damper regarding the CO2 concentration. And this one is the return that copy the supply damper command in order to extract the same amount of air that the supply bring.

tasodorff: ok. please rename RDM to RDC and its good to go.

tasodorff (changes requested):
Rename RDM to RDC and we are good to go.

tasodorff:
@ghairapetian please make recommended changes and address conflicts so this can be merged.

cstirdivant (approved):

mschulze17:
@cstirdivant yours to merge :)

PR #581, opened 2022-04-28
Remove `apparentpower` and `reactivepower`

tasodorff:
There is already precedent for apparent and reactive power where `apparent` and `reactive` are measurement_descriptor subfields. Removing these two subfields since they deviate from precedent, and the current precedent is technically correct (since power can be measured in volt-amps already); therefore it should be kept.


mschulze17:
@charbull what's the latest with this one? still waiting for Keith's review?

tasodorff:
@charbull @berkoben review this.

charbull (approved):

charbull:
please wait for @rniemo-g review

cstirdivant:
Is this good to go @tasodorff ? Or do you need Keith's approval, as well?

PR #583, opened 2022-04-28
Match mappings on raw field name

ahemphill:


tools/scoring/score/dimensions/dimension.py
```diff
@@ -360,6 +361,20 @@ def _match_virtual_entities(
 
     return matches_virtual
 
+  @staticmethod
+  def _isolate_mappings(translations, *,
+                        mapping_type: MappingType) -> Set[Tuple[str, Tuple]]:
+    """Distills mappings from each entity into a set for global comparison"""
+    mappings = set()
+    for translation in translations:
+      field = translation[1]
```
charbull, line 370: I am not sure I follow here what we are doing?

ahemphill: It does the same thing it was doing previously: takes the list of translations and returns a set of mappings ```[(standard_field_name, field )] -> {(raw_field_name, field.mappings.mapping_1), (raw_field_name, field.mappings.mapping_2)}```
The only difference is that it used to output the standard field name instead
```{(standard_field_name, field.mappings.mapping_1), (standard_field_name, field.mappings.mapping_2)}```

charbull (approved):

PR #590, opened 2022-05-10
Addition of a non standard VAV type.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -472,4 +472,14 @@ VAV_RD_CO2C:
   implements:
   - VAV
   - RD
-  - CO2C
\ No newline at end of file
+  - CO2C
+
+# Additional type for FR-PAR-50A
+VAV_HWZTC_CHWZTC_DTM:
```
tasodorff, line 478: You have listed this as canonical, but the VAV is missing supply damper and flow control. What do you mean by external? As in, it is not integrated? 

If that is the case then the appropriate thing to do would be to add SD and create this as `VAV_SD_DTM_HWZTC_CHWZTC` and implement `MISSING` fields in the translation. This describes more accurately what it does (since it _is_ a typical VAV -- it just doesn't send all the typical data).

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -472,4 +472,14 @@ VAV_RD_CO2C:
...
+  description: "Zone temperature control VAV with discharge temperature monitoring and external air volume control (not included)."
+  is_canonical: true
+  implements:
+    - VAV
+    - HWZTC
+    - CHWZTC
```
tasodorff, line 484: Please confirm: these VAVs have both heating and chilled water coils?

tasodorff (changes requested):

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
ghairapetian: Actually, this equipement does not have a damper. A DOAS bring air into this equipement for cooling or heating and that air go directly into the room. Between the DOAS and this equipement, there is a mechanical equipement that maintain the flow of air under a maximum (not by a regulation system but mechanically).
So it's not just a translation missing fields as there is no damper.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
ghairapetian: Yes I confirm that.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
tasodorff: I see. Please adjust the description to something like this: 
"A zone VAV with heating and cooling control, but no flow control; part of DOAS system where air volume is not measured or controlled locally, but instead controlled by the upstream equipment."

tasodorff: Make this change and I will approve the PR.

ghairapetian: Done

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -472,4 +472,14 @@ VAV_RD_CO2C:
   implements:
   - VAV
   - RD
-  - CO2C
\ No newline at end of file
+  - CO2C
+
+# Additional type for FR-PAR-50A
```
tasodorff, line 477: Please remove this bit. No building specific info in the ontology please.

tasodorff (approved):
Approving pending removal of site-specific comment. @ghairapetian please adjust.

PR #617, opened 2022-06-06
Updating documentation

trav3711:
Updating tools and instance validator READMEs to instruct users to run the guid generator before running the instance validator.


trav3711:
I'm not sure why checks are failing on parts of the validator that I have not touched...

PR #621, opened 2022-06-07
Addition of abstracts and canonical types for DSU (Dry scrubber unit)

RathishKJ:


ontology/yaml/resources/HVAC/entity_types/DSU.yaml
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+########################
+### Canonical Types  ###
+########################
+
+DSU_FDPM:
```
tasodorff, line 19: needs SS. Will not accept a device with just this particular field. Is there variable speed fan control?

ontology/yaml/resources/HVAC/entity_types/GENERALTYPES.yaml
```diff
@@ -37,6 +37,12 @@ FAN:
   implements:
   - EQUIPMENT
 
+DSU:
```
tasodorff, line 40: Is this a PCU? If not, how is it distinguished from a PCU? This name is not defined on the BDNS or BOS Abbreviation standard.

tasodorff (changes requested):
See comments.

ontology/yaml/resources/HVAC/entity_types/GENERALTYPES.yaml
RathishKJ: It is a type of a PCU (pollution control Unit) which is takes of the particulate matters form the exhaust air before it is left outside. Since it is a dry scrubber, we have requested for inclusion of the name under BDNS pull request 112.

RathishKJ:
Closing the PR as the dry scrubber is modeled along with the Kitchen Exhaust fan

PR #635, opened 2022-06-09
Abstracts and canonical for Elevators

RathishKJ:


RathishKJ:
> This needs to be in a new namespace, since it isnt electrical equipment. Please create a new namespace called TRANSPORT and put it in there.

The New namespace TRANSPORT has been created under which, the Abstracts, general types and canonical has been defined for Elevator

PR #642, opened 2022-06-13
Water Meter with both fields

darshanjain12:
Water meter with water_volume_accumulator and supply_water_flowrate_sensor


ontology/yaml/resources/METERS/entity_types/WM.yaml
```diff
@@ -20,6 +20,9 @@ WM_STANDARD:
   is_canonical: true
   uses:
   - water_volume_accumulator
+  opt_uses:
+  - failed_alarm
+  - supply_water_flowrate_sensor
```
tasodorff, line 25: does it need to be supply or can this just be `flowrate_sensor`?

darshanjain12: Yes it can be .
Thank you .
Updated the changes.

mschulze17:
@tasodorff back to you

tasodorff (approved):

PR #648, opened 2022-06-16
Abstract_update_AHU_Trip_status

Karthi8883:
Added trip status on basic combination of supply fan


tasodorff:
fix validator errors. if github does not run them automatically, you can still run them locally. ensure your PR is valid before review, otherwise it will inhibit feedback.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -167,6 +167,12 @@ literals:
   - ON
   - OFF
   - UNKNOWN
+- supply_fan_run_mode:
+  - AUTO
+  - MANUAL
+- bypass_supply_fan_status:
```
tasodorff, line 173: what is this, the bypass for the inverter? would this not be covered by the [`supply_fan_run_mode`](https://github.com/google/digitalbuildings/blob/e123967128248b896d03dbc24d468e7a27d066d5/ontology/yaml/resources/fields/telemetry_fields.yaml#L1513)?

or perhaps it is an inverter mode of some kind, since it seems like you are suggesting this is for the inverter itself. not sure how you would bypass a supply fan otherwise.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -381,6 +381,8 @@ SFVSC:
   - supply_fan_speed_percentage_sensor
   - supply_fan_current_sensor
   - supply_fan_power_sensor
+  - supply_fan_run_mode
```
tasodorff, line 384: not sure i would want this field on this type, given run_mode states are {AUTO, MANUAL} and probably are more related to the running of the fan than the VFD. please explain what you intend to do with this data.

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: This only bypasses the drive in case of failure and motor runs at full speed. This field shows whether the motor is running in VFD control or is bypassed. This is not covered by the supply_fan_run_mode.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Karthi8883: Updated in SFSS

tasodorff (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: Updated

tasodorff (approved):

mschulze17:
@cstirdivant all yours to merge

PR #650, opened 2022-06-16
update abstarct supply rh sensor

Karthi8883:
update abstarct supply rh sensor


tasodorff:
validate this and see comments

Karthi8883:
Updated as part of [https://github.com/google/digitalbuildings/pull/651](url)

PR #651, opened 2022-06-16
Update ABSTRACT.yaml

Karthi8883:


tasodorff:
fix validator errors

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -117,6 +117,7 @@ ZTM:
   is_abstract: true
   opt_uses:
   - discharge_air_temperature_sensor
+  - zone_air_relative_humidity_sensor
```
tasodorff, line 120: there is a type called `ZHM` that already defines this. please use it explicitly.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -5028,6 +5031,7 @@ SSPC2X:
   - supply_air_damper_percentage_command
   - supply_air_flowrate_sensor
   - supply_fan_speed_percentage_command
+  - average_supply_air_static_pressure_sensor
```
tasodorff, line 5034: what is the purpose of having this? is it used in control of the device? if so, how?

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
   - EFSS
   - HTZC
 
+AHU_SFSS_SFVSC_CHWSC_RTM_FDPM2X_ESPC_CO2DFVSC_SSPC2X_CO2M6X_ZTM_VOCM_PMM:
+  description: "Single zone AHU."
+  is_canonical: true
+  implements:
+  - AHU
+  - SFSS
```
tasodorff, line 521: `DFSS`

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
...
+  - SFVSC
```
tasodorff, line 522: `DFVSC`

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
...
+  - CHWSC
```
tasodorff, line 523: anywhere you are using supply should be changed to discharge, if this is truly a single-zone AHU

tasodorff: so in this case, `CHWDC`

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
...
+  - RTM
+  - FDPM2X
+  - ESPC
+  - CO2DFVSC
+  - SSPC2X
```
tasodorff, line 528: Why would a single-zone unit control to static pressure? This doesn't make sense on this unit. Justify.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
...
+  is_canonical: true
+  implements:
+  - AHU
+  - SFSS
+  - SFVSC
+  - CHWSC
+  - RTM
+  - FDPM2X
+  - ESPC
+  - CO2DFVSC
```
tasodorff, line 527: This conflicts with `SSPC2X` below. Does the device control to CO2 levels, or does it control to static pressure?

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
...
+  - SSPC2X
+  - CO2M6X
```
tasodorff, line 529: This is a single zone device, so why does it have 6 carbon dioxide sensors? Remove.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,24 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
   - EFSS
   - HTZC
 
+AHU_SFSS_SFVSC_CHWSC_RTM_FDPM2X_ESPC_CO2DFVSC_SSPC2X_CO2M6X_ZTM_VOCM_PMM:
```
tasodorff, line 516: this device makes no mention of outside air. im expecting an outside air damper on this. where is it? even if the device does not send that data, i still expect it on the model (and for the translation to be incomplete in the building config file). please address.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Karthi8883: It will not be used for control. It is only shown for indication in BOS. hence captured in opt_uses.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Karthi8883: Sorry, this is is a multi zone AHU, hence the usage of supply.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Karthi8883: Multi zone device.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Karthi8883: Added OAFC - will remove while translation.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Karthi8883: This controls to static Pressure but also monitors CO2 in addition.

tasodorff: well, is it supply control then? `CO2DFVSC` is not appropriate, would need a new type called `CO2SFVSC`.

tasodorff: you may also need to clarify which zone air co2 sensor is referenced by the setpoint. will this utilize an averaging sensor? a max sensor?

tasodorff (changes requested):

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Karthi8883: Updated CO2EFVSC as exhaust fan speed is controlled by the CO2. Max of zone sensor is used for zone air co2.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -513,6 +513,26 @@ AHU_DFSS_DSP_DX2ZC_ECONZ_EFSS_HTZC:
...
+  - SFSS
+  - EFSS
+  - SFVSC
+  - CHWSC
+  - RTM
+  - FDPM2X
+  - CO2EFVSC
+  - SSPC2X
+  - CO2M6X
+  - ZTM
```
tasodorff, line 530: Which zone then? if we have 6 zones, i would expect something like a ZTM6X abstract type, same for ZHM6X

tasodorff (changes requested):

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Karthi8883: Sorry, removed ZTM, ZHM as these were not part of the AHU.

tasodorff (approved):

mschulze17:
@cstirdivant yours to merge

PR #656, opened 2022-06-17
update abstract_vav

Karthi8883:
updated standard vav required points


tasodorff:
validator errors

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -30,6 +30,12 @@ SD:
   - supply_air_heating_flowrate_capacity
   - cooling_thermal_power_capacity
   - supply_air_temperature_sensor
+  - zone_air_occupied_cooling_temperature_setpoint
```
tasodorff, line 33: These do not belong here. Create an abstract type for this information, it has nothing to do with airflow control, which is the purpose of the type.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -30,6 +30,12 @@ SD:
...
+  - zone_air_unoccupied_cooling_temperature_setpoint
+  - zone_air_temperature_setpoint
+  - min_supply_air_flowrate_sensor
```
tasodorff, line 36: the min of what? min and max are aggregations across multiple sensors.

ontology/yaml/resources/METERS/entity_types/WM.yaml
```diff
@@ -20,7 +20,6 @@ WM_STANDARD:
   is_canonical: true
   uses:
   - water_volume_accumulator
-  opt_uses:
```
tasodorff, line 23: not allowed, this will break existing types. must remain optional.

tasodorff (changes requested):
see comments

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Karthi8883: This is the min and max air flow rate values between which the vav can operate.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Karthi8883: Updated

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
tasodorff: high limit and low limit is what you should use then. see [`HLSAFS`](https://github.com/google/digitalbuildings/blob/accdc3b2685b08531cd83fcdd048f432bb2a2008/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml#L4557) for an example.

tasodorff: it also bounds the setpoint, not the sensor. looks like this might be a mistake here.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -148,6 +151,16 @@ CSP:
   - zone_air_temperature_sensor
   - zone_air_cooling_temperature_setpoint
 
+CSPO:
+  description: "Single cooling setpoint control with occupancy."
```
tasodorff, line 155: you do not have single cooling setpoint control, you have two and one is chosen based on occupancy. please adjust description.

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
```diff
@@ -39,6 +39,15 @@ VAV_SD_DSP:
   - SD
   - DSP
 
+VAV_SD_ZTC_CSPO:
+  description: "Simple cooling only VAV with occupancy."
+  is_canonical: true
+  implements:
+  - VAV
+  - SD
+  - ZTC
```
tasodorff, line 48: ZTC includes `zone_air_temperature_setpoint`, and your other type defines `zone_air_occupied_cooling_temperature_setpoint` and `zone_air_unoccupied_cooling_temperature_setpoint`.  We shouldnt have all of these setpoints. If you end up with an effective zone temp setpoint, like `zone_air_cooling_temperature_setpoint` then that would be the only one we need. You could then change `CSP` to include `occupancy_status` as an optional field and use that.

tasodorff: but if you want both setpoints, then we would have [this type](https://github.com/google/digitalbuildings/blob/accdc3b2685b08531cd83fcdd048f432bb2a2008/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml#L639) you could reference. Update it for cooling and add occupancy status as required (and keep the naming convention intact).

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -453,6 +453,10 @@ literals:
 - zone_air_static_pressure_setpoint
 - zone_air_voc_concentration_sensor
 - zone_air_voc_concentration_setpoint
+- min_supply_air_flowrate_sensor
```
tasodorff, line 456: remove

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -453,6 +453,10 @@ literals:
...
+- max_supply_air_flowrate_sensor
```
tasodorff, line 457: remove

tasodorff (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: Done

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: Done

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Karthi8883: Modified and used HLSAFS

ontology/yaml/resources/HVAC/entity_types/VAV.yaml
Karthi8883: We do have an effective set point and hence have modeled as per your suggestion and added occupancy status as an optional field in CSP.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -453,6 +453,8 @@ literals:
 - zone_air_static_pressure_setpoint
 - zone_air_voc_concentration_sensor
 - zone_air_voc_concentration_setpoint
+- zone_air_occupied_cooling_temperature_setpoint
```
tasodorff, line 456: these appear no longer used in this PR. please remove.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -453,6 +453,8 @@ literals:
...
+- zone_air_unoccupied_cooling_temperature_setpoint
```
tasodorff, line 457: same

tasodorff (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: Updated

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: Removed

tasodorff (approved):

mschulze17:
@cstirdivant yours to merge

PR #662, opened 2022-06-19
CRAC Unit  with fields and Canonical type added in AHU

darshanjain12:


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -3601,6 +3601,21 @@ AHU_DFSS_DXZC_ECONZ_EFSS2X_HT2ZC_CO2C:
   - HT2ZC
   - CO2C
 
+AHU_DFSS_DXRC_RHDHC_AL:
```
tasodorff, line 3604: does not contain return or outside air control, required for AHUs.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -729,7 +729,16 @@ literals:
 - lost_power_alarm:
   - ACTIVE
   - INACTIVE
-
+- refrigerant_failed_alarm:
+  - ACTIVE
+  - INACTIVE
+- outside_heating_exchange_failed_alarm:
```
tasodorff, line 735: outside a heat exhanger? what does this mean exactly?

darshanjain12: It is coil responsible for thermal energy exchange failed alarm. Can it be a heat_exchange_failed_alarm .
Done the commit

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
darshanjain12: Yes i think its my mistake to add it in AHU adding it in FCU. Does that works?

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -743,7 +743,16 @@ literals:
 - lost_power_alarm:
   - ACTIVE
   - INACTIVE
-
+- refrigerant_failed_alarm:
```
cstirdivant, line 746: shoud this be a `refrigerant_leak_alarm` ?

cstirdivant (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
darshanjain12: Actually this is a refrigerant sensor fault alarm

darshanjain12: > shoud this be a `refrigerant_leak_alarm` ?
Actually this is a refrigerant sensor fault alarm

darshanjain12: Removing the refrigerant sensor failed alarm as cannot model it because sensor is a point type and cannot be used  as descriptor.

cstirdivant (approved):

PR #664, opened 2022-06-19
Fields for heat meter added

darshanjain12:
Field for Δt calculated


ontology/yaml/resources/METERS/entity_types/HM.yaml
```diff
@@ -48,6 +48,9 @@ HM_CHW_WAC:
   is_canonical: true
   uses:
   - water_volume_accumulator
+  opt_uses:
+  - exchange_thermal_temperature_sensor
```
tasodorff, line 52: What is the context for this? exhange thermal?

darshanjain12: It is actually a temperature difference between supply water and return water in chilled water pipeline.
Can it be differential temperature sensor

darshanjain12: Did the changes

ontology/yaml/resources/METERS/entity_types/HM.yaml
```diff
@@ -48,6 +48,9 @@ HM_CHW_WAC:
   is_canonical: true
   uses:
   - water_volume_accumulator
+  opt_uses:
```
cstirdivant, line 51: Please put these `opt_uses` on the original `HM_CHW` type

cstirdivant (changes requested):

ontology/yaml/resources/METERS/entity_types/HM.yaml
cstirdivant: please remove from here now, it is redundant

cstirdivant (changes requested):

ontology/yaml/resources/METERS/entity_types/HM.yaml
```diff
@@ -39,6 +39,10 @@ HM_CHW:
   - chilled_return_water_temperature_sensor
   - cooling_thermal_power_sensor
   - cooling_thermal_energy_accumulator
+  opt_uses:
+  - differential_temperature_sensor
+  - failed_alarm
+  - water_volume_accumulator
```
cstirdivant, line 45: remove this field, since we have `HM_CHW_WAC` that uses this as optional

cstirdivant (changes requested):

cstirdivant (approved):

PR #670, opened 2022-06-22
Update FD.yaml

Karthi8883:
Points are required to monitor supply and return dampers status


tasodorff:
validator errors

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -161,6 +161,21 @@ literals:
 - supply_air_temperature_setpoint
 - supply_fan_current_sensor
 - supply_fan_power_sensor
+- supply_air_damper_open_status:
```
cstirdivant, line 164: is it possible to model this as one field `supply_air_damper_status` ? or are there legitimately two separate fields for open and closed?

cstirdivant (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
Karthi8883: There are two fields. The status in Open field will be active only when the damper is fully open and the status in the Closed field will be active only when the damper is fully closed. If the damper is for example is stuck in an intermediate position, neither of the fields will be active.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -161,6 +161,21 @@ literals:
...
+- supply_air_damper_closed_status:
+  - OPEN
+  - CLOSED
+- return_air_damper_open_status:
+  - OPEN
+  - CLOSED
+- return_air_damper_closed_status:
+  - OPEN
+  - CLOSED
+- high_temperature_alarm:
```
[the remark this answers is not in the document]
Karthi8883: There is a sensor placed in the zone where the panel is located, this field is active when the sensor reads a high value. Can we model it as zone_high_temperature_alarm?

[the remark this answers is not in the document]
tasodorff: should be `high_zone_air_temperature_alarm` in that case. there are other high temp alarms that follow that precedent.

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
```diff
@@ -19,4 +19,11 @@ FD_STD:
   implements:
   - FD
   uses:
-  - damper_status
\ No newline at end of file
+  - damper_status
+  opt_uses:
+  - supply_air_damper_open_status
```
tasodorff, line 24: if there is a supply air damper on the same device, then what is the damper above? either remove these from this FD or differentiate the field above.

tasodorff (changes requested):

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
Karthi8883: This field was already existing. Did not want to break the existing field, hence modeled it in opt_uses with both supply and return dampers.

ontology/yaml/resources/fields/telemetry_fields.yaml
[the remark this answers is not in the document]
Karthi8883: Updated

ontology/yaml/resources/SAFETY/entity_types/FD.yaml
Karthi8883: The field was already existing. Did not want to break existing model, hence modeled under opt_uses for both supply and return dampers

tasodorff: but the supply and return on what? typically a fire smoke damper is its own stand-alone damper, and there is no reference to supply or return air.

tasodorff: i think you are modeling too many dampers as part of the FD.

Karthi8883: Noted, we will close this pull request and model a generic FD with two dampers as a separate pull request.

PR #676, opened 2022-06-22
Update Facilities.yaml

RathishKJ:


RathishKJ:
Used the zone defined in HVAC namespace. Therefore deleted the current pull request.

PR #678, opened 2022-06-23
Update WM.yaml

Karthi8883:
added water_linearvelocity_sensor point


tasodorff:
validator errors

ontology/yaml/resources/METERS/entity_types/WM.yaml
```diff
@@ -31,4 +31,7 @@ WM_WF:
   - WM
   is_canonical: true
   uses:
-  - supply_water_flowrate_sensor
\ No newline at end of file
+  - supply_water_flowrate_sensor
+  opt_uses:
+  - water_linearvelocity_sensor
```
tasodorff, line 36: if we are being specific about it being supply, then please add supply

tasodorff (changes requested):

ontology/yaml/resources/METERS/entity_types/WM.yaml
Karthi8883: We are not specific about it being supply - prefer it to be generic but the supply_water_flowrate_sensor field was already existing and did not want to break existing model and hence added new fields in the opt_uses.

tasodorff (approved):
Opening issue to remove `supply` from this definition.

mschulze17:
@cstirdivant yours to merge

PR #695, opened 2022-06-27
Update_freezer

Karthi8883:
updated freezer monitoring points


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -192,6 +192,18 @@ DX2RC:
...
+FRZM:
+  description: "Freezer monitoring."
+  is_abstract: true
+  opt_uses:
+   - failed_alarm
+   - run_status
+   - compressor_run_status
+   - run_time_accumulator
+  uses:
+   - temperature_sensor
```
tasodorff, line 204: what temperature?

tasodorff (changes requested):
model these after other chiller types. where is the start command? water temps? is this actually a chiller?

see precedent in the CH.yaml file.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Karthi8883: This is a freezer used in kitchen. Will remove this canonical type from chiller and add it to the CRREF.yaml approved yesterday. The temperature is the freezer temperature.

Karthi8883:
Will create new pull request based on recently approved pull requests.

PR #698, opened 2022-06-29
Added Individual DX speeds

tom-skoczylas:
@mschulze17 @tasodorff @tsodorff


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -1508,6 +1508,8 @@ DX2ZC:
   - cooling_thermal_power_capacity
   - cooling_percentage_command
   - compressor_speed_percentage_command
+  - compressor_speed_percentage_command_1
```
tasodorff, line 1511: @cstirdivant thoughts on this? does it make sense to have the enums and non enums in the same type?

tasodorff:
@berkoben @charbull @trav3711 @sypks the abstract type defines enum and non enum for the same type:
- `speed_percentage_command`
- `speed_percentage_command_1`
- `speed_percentage_command_2`

The validator throws no error. I think this needs to be fixed; I don't think it should be valid for the same type to have both enums and non enums. Thoughts?

PR #703, opened 2022-07-02
Walkin room Freezer

darshanjain12:


ontology/yaml/resources/HVAC/entity_types/FRZ.yaml
```diff
@@ -28,6 +28,15 @@ FRZ_DXZTC_DFSS:
   - DXZTC
   - DFSS
 
-
-
+FRZ_DFSS_DEFSS_DEFSTC:
+  description: "Freezer with discharge fan,compressor, defrost run and temperature control."
```
cstirdivant, line 32: you mention "compressor" but I do not see any compressor types here.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FRZ.yaml
darshanjain12: Yes not included thank you for noticing added the required abstract type.

cstirdivant (approved):

PR #713, opened 2022-07-11
SENSORS NEW CANONICAL TYPE

darshanjain12:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -401,6 +401,7 @@ SFVSC:
   - supply_fan_failed_alarm
   - control_mode
   - supply_fan_power_status
+  - master_alarm
```
tasodorff, line 404: Two comments:
1. Is this alarm name appropriate? Is there a more specific name that can be used.
2. This is not appropriate for the VFD abstract type. Why is this not defined on the canonical type?

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -424,6 +425,7 @@ DFVSC:
   - discharge_fan_energy_accumulator
   - discharge_fan_failed_alarm
   - control_mode
+  - master_alarm
```
tasodorff, line 428: same as above.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -5326,6 +5328,15 @@ PMM:
   implements:
   - MONITORING
 
+PMM2X:
+  description: "Particulate matter monitoring for 2 zones."
```
tasodorff, line 5332: The fields you have defined describe two separate sensors of different types, not the same type of sensor duplicated across multiple zones. For example, this is what i would expect if it was two zones:
```
uses:
- zone_air_pm2pt5_concentration_sensor_1
- zone_air_pm2pt5_concentration_sensor_2
```

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
```diff
@@ -233,5 +233,15 @@ SENSOR_ZA_CO2M_PMM_HCHO:
   - CO2M
   - PMM
   uses:
-  - zone_air_formaldehyde_concentration_sensor  
-  
\ No newline at end of file
+  - zone_air_formaldehyde_concentration_sensor
+
+SENSOR_PMM2X_CO2M13X:
```
tasodorff, line 238: I do not understand why these would be modeled together. Instead, model each individual zone CO2 sensor and then connect them to the floor in building config. If you are doing this because you have a device that is proxying this data for you (e.g. a sensor gateway) then use the `GATEWAYS/PASSTHROUGH` entity type, so you dont have to define a bespoke, one-off entity type. Then, using links in Building Config, you can link the gateway to the individual sensor models you create.

tasodorff (changes requested):
Please see comments.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
darshanjain12: 1. Here in our context the alarm is an master alarm for example if  any alarm is active then the master alarm will be active.
2. Will add that as part of canonical type.

ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml
darshanjain12: Actually this all 13 CO2 sensors , 2 PM2.5 Sensors are inside a single area served by a single AHU.

tasodorff: your description for `CO2M13X` designates these as 13 separate zones. are these really all placed in one control zone?

PR #714, opened 2022-07-12
Lighting canonical type

darshanjain12:


ontology/yaml/resources/LIGHTING/entity_types/LGRP.yaml
```diff
@@ -225,6 +225,19 @@ LGRP_OS_IM_LSC_LC_PWM_SCZ:
...
+  description: "Standard Lighting fixture group with occupancy based,area switch status and energy monitoring."
+  is_canonical: true
+  implements:
+  - LGRP
+  - OS
+  - LSC
+  - ENM
+  opt_uses:
+  - total_lamp_failed_counter
+  - wireless_device_failed_counter
```
tasodorff, line 238: Is this a `total` as well?

ontology/yaml/resources/LIGHTING/entity_types/LGRP.yaml
```diff
@@ -225,6 +225,19 @@ LGRP_OS_IM_LSC_LC_PWM_SCZ:
...
+  - communication_failed_counter
```
tasodorff, line 239: Is this meaning that the lighting group is not communicating, or that the devices within the lighting group arent communicating? If the latter, please use `total` as well.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -379,6 +379,7 @@ literals:
   - OCCUPIED
   - UNKNOWN
   - UNOCCUPIED
+  - INACTIVE
```
tasodorff, line 382: Whats the difference between `UNOCCUPIED` and `INACTIVE`?

tasodorff (commented):
Please see questions.

ontology/yaml/resources/LIGHTING/entity_types/LGRP.yaml
darshanjain12: Yes it is the latter one. Wireless device failed counter is the total number of  wireless devices failed in the lighting group and communication failed counter is the total number of devices are not able to communicate in the lighting group.
PS: Added the total for both the fields.

ontology/yaml/resources/LIGHTING/entity_types/LGRP.yaml
darshanjain12: Yes

ontology/yaml/resources/fields/telemetry_fields.yaml
darshanjain12: UNKNOWN: Unknown is state which is caused due to some kind of fault/error example: battery fault

INACTIVE: It is the state which indicates that the sensor output is not taken into consideration. example for a particular area and for particular time schedule the sensor output is not taken into consideration and the lights are always turned ON or OFF.

tasodorff: these seem mutually exclusive. one applies to the state of the zone (occupied or unoccupied) and one applies to the light itself (run_status). i think these should be separated and INACTIVE not added as a state to the zone_occupancy_status.

darshanjain12: This "INACTIVE" state is for the zone itself which means that the irrespective of the output as a state which can be "OCCUPIED" and "UNOCCUPIED" is not taken into consideration for that particular zone.

tasodorff: but shouldnt these be separate fields? i expect the `zone_occupancy_status` to tell me whether the zone is occupied or unoccupied, and thats it. you cannot override a zone status. if this were a `zone_command` (or something similar, i dont believe this is defined yet), then your states might make more sense... the zone is commanded to occupy, unoccupy, or inactive. i would still argue this is not a correct representation because the zone can either be commanded active or inactive -- occupied is active and inactive/unoccupy represent the same thing (inactive).

is it possible to separate this functionality into two fields?

darshanjain12: Hi Trevor,

Something is creating confusion. To make you understand the functionality i have attached the snapshot of the state description provided by the vendor I believe this will help. For more detailed understanding attaching the document too.

[https://www.lutron.com/TechnicalDocumentLibrary/369998.pdf]
(image)

darshanjain12: Done the required changes

tasodorff: yes but you added it below on line 399. good

tasodorff (approved):

mschulze17:
@cstirdivant yours to merge

PR #715, opened 2022-07-14
Make service accounts optional, falling back to gcloud application default credentials

noursaidi:
Removes the need for service accounts, using gcloud application default credentials when a service account is not provided


tools/validators/instance_validator/README.md
```diff
@@ -41,15 +41,16 @@ Note: as of the current development stage, you must clone the entire repository
...
 
-If you would like to use the telemetry validation mode, you must provide the "--subscription" and "--service-account" parameters when running instance_validator.py. Failure to provide both of these parameters will result in early termination of the validator and an error message. If you do not provide either parameter, the validator will run with telemetry validation mode disabled.
+If you would like to use the telemetry validation mode, you must provide the "--subscription", and you must either be authentication with gcloud application-default credentials, or provide a "--service-account" parameters when running instance_validator.py. Failure to provide  these parameters will result in early termination of the validator and an error message. If you do not provide either parameter, the validator will run with telemetry validation mode disabled.
```
charbull, line 44: can you please add something like `with the adequate google project id provided by your google point of contact` or something in those lines?

tools/validators/instance_validator/validate/subscriber.py
```diff
@@ -19,7 +19,7 @@
 from concurrent import futures
 import json
 
-from google.auth import jwt
+import google.auth
```
charbull, line 22: can you please keep the previous form? our internal linter will not accept the new suggestion

tools/validators/instance_validator/validate/subscriber.py
```diff
@@ -31,10 +31,9 @@ class Subscriber(object):
     service_account_info: service account information from the GCP project.
```
charbull, line 31: can you please add `(optional) when not provided, defaults to gcloud credentials` or similar

tools/validators/instance_validator/validate/subscriber.py
```diff
@@ -31,10 +31,9 @@ class Subscriber(object):
...
   """
 
-  def __init__(self, subscription_name, service_account_info_json_file):
+  def __init__(self, subscription_name, service_account_info_json_file = None):
```
charbull, line 34: can you please update this to : `service_account_info_json_file : Optional [str] =None`

charbull (commented):
Hi Nour, 
thank you for the pull request.
minor suggestions.

Can you please sign the cla :)

tools/validators/instance_validator/README.md
noursaidi: Interesting, didn't capture this in testing but I think I need to add an additional parameter for the project , because normal user accounts can't set the google projects as the default project. So either the full subscription path `projects/<PROJECT_ID>/subscriptions/<SUBSCRIPTION_NAME>`needs to be given or an additional parameter `project_id`. Will update the PR to incorporate this for review

noursaidi: Actually I understand current guidance is putting the full subscription path so no further changes needed

charbull: [sorry for the delay, just seeing this, please trigger the review when you are ready :) ]

correct but the user must be have the correct id with the gcloud auth otherwise they will get an error of unauthentication. what do you think?

charbull (commented):
[sorry for the delay, just seeing this, please trigger the review when you are ready :) ]

charbull:
@noursaidi can you handle the conflicts in the README.md please :)

tools/validators/instance_validator/README.md
noursaidi: Sorry just picking this up again. I don't think it's needed for actually, I've tested on my machine by removing gcloud credentials and settings, and only setting up the application-default credentials and it worked for me. I want to test on a fresh install to confirm though, so will post back.

tools/validators/instance_validator/validate/subscriber.py
```diff
@@ -18,23 +18,28 @@
 
 from concurrent import futures
 import json
+import google.auth
```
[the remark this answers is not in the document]
noursaidi: Sorry didn't complete the changes yet but now have, PTAL. I now `import auth from google` and use `auth.jwt`

noursaidi:
@charbull PTAL, comments should all now be incorporated.

I've updated the documentation to mention `gcloud init` as suggested. I've tested it on a fresh Debian install and only application-default is needed for this change, but it would be better to have users have gcloud fully setup too.

charbull (approved):

charbull:
@noursaidi approved, there are github actions failure due to linter nits:
https://github.com/google/digitalbuildings/runs/7439781436?check_suite_focus=true

Thank you for the contribution :)

noursaidi:
Ah, Thanks for taking a look. I saw that, but assumed it would be ignored because it was an existing long line. 

Nnow shortened, but it's not consistent with how long lines are wrapped elsewhere in the code. Let me know if that's fine

charbull:
@noursaidi can you please take a look. There are some tests failing

charbull (changes requested):
@noursaidi can you check the tests please some are failing.

thank you

noursaidi:
Identified the issue - changing `if bool(subscription) != bool(service_account):` to `if not bool(subscription)` causing issues with failure of CI testing due to missing subscription. Just working my way through it

noursaidi:
@charbull The logic `if bool(subscription) != bool(service_account):` would have the validator exit if only one of the subscription or service account was provided. If neither was provided, the validator continues, does stuff, and if a subscription is provided starts telemetry validation, otherwise writes the reports and stops.

I've removed the logic as I don't think it's applicable. Then, the new behaviour matches the existing behaviour just without the early exit. 

I've also removed the two CI tests which test for only passing a service account and only passing a subscription, as these tested for exit and I couldn't think what the alternative test is. There is tests for passing subscriptions and/or service accounts, and equally without, so I don't think there's any reduction in test coverage

charbull (approved):

PR #723, opened 2022-07-21
ignore state messages in telemetry validation

noursaidi:
Posting PR for testing/feedback purposes - we seem to have an issue that messages which aren't telemetry are being validated as telemetry, resulting `Field missing from telemetry message`  errors for each field for an entity. This seems to have fixed it, but I still need to investigate if this is the root cause.

This uses the [sub blocks from UDMI](https://github.com/faucetsdn/udmi/blob/master/docs/specs/subblocks.md) to ignore anything which is a state message or not `pointset`


noursaidi:
Does the instance validator check units, which in UDMI are only in the state messages?

charbull:
> Does the instance validator check units, which in UDMI are only in the state messages?

Hi @noursaidi the instance validator does not read from state messages yet. 
The state validation is here currently: https://github.com/google/digitalbuildings/blob/master/tools/validators/instance_validator/validate/telemetry_validator.py#L155

noursaidi:
Thanks @charbull, can we have a message filter or similar added to to ignore messages which aren't intended to be telemetry?

This PR gives an example of a filter based on UDMI. 

This would solve the problem we're having where the validator gives an error of "Field missing from telemetry message" when it sees non-telemetry messages first.

charbull:
@noursaidi do you want to add a `--udmi` option for the telemetry validation that does what you want?

pisuke:
> @noursaidi do you want to add a `--udmi` option for the telemetry validation that does what you want?

As long as we can avoid the validator raising errors for state messages, this would work. 
But this is not only UDMI specific, shouldn't the validator only validate events messages anyway?

noursaidi:
@charbull @pisuke PTAL

noursaidi:
Fixing CI errors first, will tag again, sorry!

tools/validators/instance_validator/tests/message_filters_test.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2020 Google LLC
```
charbull, line 1: nit 2022

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -194,7 +194,7 @@ class TelemetryValidatorTest(absltest.TestCase):
 
   def testTelemetryValidatorGetErrorsReturnsAllErrors(self):
     validator = telemetry_validator.TelemetryValidator(GOOD_ENTITIES_1, 1,
-                                                       _NullCallback)
+                                                       False, _NullCallback)
```
charbull, line 197: can you please set `isUdmi=False` here and in the tests below

tools/validators/instance_validator/validate/message_filters.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2020 Google LLC
```
charbull, line 1: nit: 2022

tools/validators/instance_validator/validate/telemetry_validator.py
```diff
@@ -33,10 +34,11 @@ class TelemetryValidator(object):
   Attributes;
     entities: a dict with entity_name as a key and EntityInstance as value.
     timeout: the max time the validator must read messages from pubsub.
+    udmi: true/false treat telemetry stream as UDMI
     callback: the method called by the pubsub listener upon receiving a msg.
   """
 
-  def __init__(self, entities, timeout, callback):
+  def __init__(self, entities, timeout, udmi, callback):
```
charbull, line 41: `is_udmi`

charbull (changes requested):
thank you Nour, just nits for the syntax

tools/validators/instance_validator/tests/message_filters_test.py
noursaidi: Changed

tools/validators/instance_validator/validate/message_filters.py
noursaidi: changed

tools/validators/instance_validator/validate/handler.py
```diff
@@ -90,10 +90,10 @@ def _ValidateConfig(
 
 def _ValidateTelemetry(subscription: str, service_account: str,
                        entities: Dict[str, entity_instance.EntityInstance],
-                       timeout: int) -> None:
+                       timeout: int, udmi: bool) -> None:
```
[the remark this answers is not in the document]
noursaidi: changed all occurrences of `udmi` to `is_udmi`

tools/validators/instance_validator/validate/telemetry_validator.py
noursaidi: changed

tools/validators/instance_validator/tests/telemetry_validator_test.py
noursaidi: I think I've made the change requested, but please take a look and let me know because I wasn't sure if I understood it properly.

tools/validators/instance_validator/tests/telemetry_validator_test.py
```diff
@@ -293,8 +300,9 @@ def testTelemetryValidatorDetectsStringAsInvalidNumber(self):
     self.assertLen(errors, 1)
 
   def testTelemetryValidatorDetectsBooleanAsInvalidNumber(self):
+    is_udmi = False
     validator = telemetry_validator.TelemetryValidator(GOOD_ENTITIES_1, 1,
-                                                       _NullCallback)
+                                                       is_udmi, _NullCallback)
```
charbull, line 305: you could do it here directly, `is_udmi=False` there is no need to declare it before.

```
    validator = telemetry_validator.TelemetryValidator(GOOD_ENTITIES_1, 1,
                                                       _NullCallback)
                                                       is_udmi=False, _NullCallback)
                                                       ```

File: tools/validators/instance_validator/validate/telemetry_validator.py [hunk 7]
```diff
@@ -33,10 +34,11 @@ class TelemetryValidator(object):
   Attributes;
     entities: a dict with entity_name as a key and EntityInstance as value.
     timeout: the max time the validator must read messages from pubsub.
+    is_udmi: true/false treat telemetry stream as UDMI
```
charbull, line 37: nit: `when true`

charbull (changes requested):
just two nits, thank you Nour

charbull (approved):
merging for now will update the nits after

PR #739, opened 2022-08-12
Pip support

charbull:
adding pip


trav3711 (commented):
Yes! I think using a requirements file with pip will make everything so much easier for the user. 

I'm in favor of totally deprecating setuptools as it's a lot more confusing and references the .egg files rather than the source code itself. setuptools is a versioning software that we're not really using for versioning.

trav3711 (approved):

PR #740, opened 2022-08-12
reactive_power to reactivepower

pisuke:
Changing the reactive power measurement name to match the unit specification


pisuke:
@charbull this PR solves inconsistencies between the `reactivepower` name in the units file and the `reactive_power` fields in several other locations. The same needs to be done with `apparentpower` but I would do it in a separate PR.

tasodorff:
we reverted this a while ago. why would we not stick with apparent and reactive as measurement descriptors?

PR #742, opened 2022-08-12
Prevent telemetry validation terminating when handling invalid payloads

noursaidi:
At the moment, a payload which comprises the number `1` or is missing required fields `timestamp` or `version` results in the validator terminating with an error. This moves a few things in the existing try/catch block to prevent this from happening, and also prevents terminating when `json.loads` doesn't return an object (because for some reason `json.loads(1)` returns 1?!

New behaviour:
```
The following Json payload is missing required field 'version'
b'{}'
The following payload is invalid:
b'1'
The following Json payload is missing required field 'timestamp'
b'{"version":"1.3.14"}'
The following payload is invalid:
b'blahblahblah'
```

### DIFF
--- a/tools/validators/instance_validator/validate/telemetry.py
+++ b/tools/validators/instance_validator/validate/telemetry.py
@@ -96,11 +96,6 @@ class Telemetry(object):
     version, timestamp, points, is_partial = (None, None, None, None)
     try:
       json_object = json.loads(message)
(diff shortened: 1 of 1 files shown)


noursaidi:
@pisuke this should fix the issue we had earlier

charbull (approved):

PR #745, opened 2022-08-13
Patching the telemetry validator to udmi errors

charbull:
Some devices are sending non conform udmi messages such as `1` or empty bytes.
Some udmi devices set the version as int for version 1. or string for version 1+


charbull:
Merging to unblock, will add unit tests in a separate PR.

PR #750, opened 2022-08-18
Doc links fix

trav3711:
One more shot at fixing this. Referring to Stack Overflow and Jekyll documentation:

https://stackoverflow.com/questions/16316311/github-pages-and-relative-paths

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,4 +1,3 @@
 theme: jekyll-theme-cayman
-
 url: https://google.github.io
 baseurl: '/digitalbuildings'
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -1,3 +1,2 @@
 theme: jekyll-theme-cayman
-url: https://google.github.io
 baseurl: '/digitalbuildings'
```
charbull, line 2: it was set to this?

charbull (approved):

PR #753, opened 2022-08-23
New EM Canonical Type

cstirdivant:
New power meter type that only has phase-level monitoring.

@tasodorff @mschulze17


tasodorff:
i will approve this if we need to, but i would push back on the availability of no meter-level KW/KWH... do we achieve this with virtual fields?

PR #764, opened 2022-09-13
Update setup.py

charbull:


charbull:
forcing it now to fix the dependencies will fix rdf generator linter separately

PR #776, opened 2022-09-21
Abel presubmit workflow

trav3711:


.github/workflows/tools.yml
```diff
@@ -7,105 +7,133 @@ on:
...
 jobs:
   build:
+    name: [ master-runner ]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@v2
```
charbull, line 25: remove this?

.github/workflows/tools.yml
```diff
@@ -7,105 +7,133 @@ on:
...
         python -m pip install pylint
+    - name: Run Pylint on tools
+      run: |
+        pylint --rcfile ./styles/pylintrc ./tools
+
 
      #TODO(b/210730979): Investigate calling ontology validator setup with instance validator setup
      #---------- Ontology Explorer ----------#
-    - name: Ontology Validator Install dependencies
-      run: |
```
charbull, line 36: why deleting those?

.github/workflows/tools.yml
```diff
@@ -7,105 +7,133 @@ on:
...
-        python -m pip install .
-      working-directory: ./tools/validators/instance_validator
     - name: Ontology Explorer Install dependencies
+      if: steps.changes.outputs.explorer == 'true'
       run: |
         python setup.py install
       working-directory: ./tools/explorer
-    - name: Run Pylint on Ontology Explorer
-      run: |
```
charbull, line 48: pylints must stay?

.github/workflows/tools.yml
```diff
@@ -7,105 +7,133 @@ on:
...
-        pylint --rcfile ./styles/pylintrc ./tools/explorer
     - name: Ontology Explorer Tests
```
charbull, line 63: this will break though? it will not find the ontology type validator dependencies

trav3711: I fixed the order of operations.

.github/workflows/tools.yml
trav3711: I fixe the order of operations, so each tool installs only its own library.

.github/workflows/tools.yml
trav3711: weird....I did't add this. Removed.

.github/workflows/tools.yml
```diff
@@ -31,81 +49,114 @@ jobs:
...
+      if: steps.changes.outputs.explorer == 'true'
+      run: |
+        pylint --rcfile ./styles/pylintrc ./tools/explorer
+    - name: Ontology Explorer Tests
+      if: steps.changes.outputs.explorer == 'true'
+      run: |
+       python -m unittest discover -p '*_test.py' --failfast -v
+      working-directory: ./tools/explorer
+
     #---------- Configuration scoring ----------#
```
charbull, line 99: nit move this last after the rdf generator it is the least that is going to get updated

.github/workflows/tools.yml
```diff
@@ -31,81 +49,114 @@ jobs:
...
     - name: Configuration scoring Install dependencies
+      if: steps.changes.outputs.scoring == 'true'
       run: |
         python setup.py install
       working-directory: ./tools/scoring
-    - name: Run Pylint on Configuration Scorer
+    - name: Run pylint on Instance Scorer
```
charbull, line 105: nit: keep the same name please: Configuration Scorer

.github/workflows/tools.yml
```diff
@@ -31,81 +49,114 @@ jobs:
...
+      run: |
+        pylint --rcfile ./styles/pylintrc ./tools/rdf_generator
     - name:  RDF Generator Tests
+      if: steps.changes.outputs.rdf_generator == 'true'
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/rdf_generator/rdfformat/tests
-    - name: Run Pylint on RDF Generator
+
+      #---------- ABEL ----------#
```
charbull, line 148: nit move abel above RDF generator?

charbull (changes requested):

.github/workflows/tools.yml
trav3711: Done.

.github/workflows/tools.yml
trav3711: Done.

.github/workflows/tools.yml
trav3711: Done. Moved scorer to last.

trav3711:
Completely removed ABEL from the workflow because tests were also having trouble.

charbull (approved):

PR #780, opened 2022-09-23
Action dependencies

trav3711:
Addressing dependency action bugs

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -73,7 +73,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/validators/instance_validator
     - name: Instance Yaml Validator Tests
-      if: steps.changes.outputs.instance_validator == 'true' OR steps.changes.outputs.type_validator == 'true'
+      if: steps.changes.outputs.instance_validator == 'true' || steps.changes.outputs.type_validator == 'true'
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/validators/instance_validator
@@ -89,7 +89,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/abel
     - name: ABEL Tests
-      if: steps.changes.outputs.abel == 'true' OR steps.changes.outputs.instance_validator == 'true' OR steps.changes.outputs.type_validator == 'true'
+      if: steps.changes.outputs.abel == 'true' || steps.changes.outputs.instance_validator == 'true' || steps.changes.outputs.type_validator == 'true'
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/abel
@@ -105,7 +105,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/guid_generator
     - name:  GUID Generator Tests
-      if: steps.changes.outputs.guid_generator == 'true' OR steps.changes.outputs.instance_validator == 'true'
+      if: steps.changes.outputs.guid_generator == 'true' || steps.changes.outputs.instance_validator == 'true'
       run: |
        python -m unittest discover -p '*_test.py' --failfast -v
       working-directory: ./tools/guid_generator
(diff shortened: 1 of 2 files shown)


tools/pip_install.sh
```diff
@@ -89,4 +89,5 @@ ontology_validator_setup
 instance_validator_setup
 explorer_setup
 guid_generator_setup
+abel_setup
```
charbull, line 92: does this add the pip in the abel setup?

.github/workflows/tools.yml
```diff
@@ -138,7 +138,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/explorer
     - name: Ontology Explorer Tests
-      if: steps.changes.outputs.explorer == 'true'
+      if: steps.changes.outputs.explorer == 'true' OR steps.changes.outputs.type_validator == 'true' OR steps.changes.outputs.instance_validator == 'true'
```
charbull, line 141: does the explorer relies on the instance validator? I dont think so

.github/workflows/tools.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/validators/instance_validator
     - name: Instance Yaml Validator Tests
-      if: steps.changes.outputs.instance_validator == 'true'
+      if: steps.changes.outputs.instance_validator == 'true' OR steps.changes.outputs.type_validator == 'true'
```
charbull, line 76: the or is not supported, please use the expression here: https://docs.github.com/en/actions/learn-github-actions/expressions#operators

.github/workflows/tools.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/abel
     - name: ABEL Tests
-      if: steps.changes.outputs.abel == 'true'
+      if: steps.changes.outputs.abel == 'true' OR steps.changes.outputs.instance_validator == 'true' OR steps.changes.outputs.type_validator == 'true'
```
charbull, line 92: the or is not supported, please use the expression here: https://docs.github.com/en/actions/learn-github-actions/expressions#operators

.github/workflows/tools.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/guid_generator
     - name:  GUID Generator Tests
-      if: steps.changes.outputs.guid_generator == 'true'
+      if: steps.changes.outputs.guid_generator == 'true' OR steps.changes.outputs.instance_validator == 'true'
```
charbull, line 108: the or is not supported, please use the expression here: https://docs.github.com/en/actions/learn-github-actions/expressions#operators

.github/workflows/tools.yml
```diff
@@ -154,7 +154,7 @@ jobs:
       run: |
         pylint --rcfile ./styles/pylintrc ./tools/scoring
     - name: Configuration Scorer Tests
-      if: steps.changes.outputs.scoring == 'true'
+      if: steps.changes.outputs.scoring == 'true' OR steps.changes.outputs.instance_validator == 'true'
```
charbull, line 157: the or is not supported, please use the expression here: https://docs.github.com/en/actions/learn-github-actions/expressions#operators

charbull (changes requested):

tools/pip_install.sh
trav3711: Yeah I don't think anymore needs to be added. I checked in the instance validator, for example, and there aren't extra files besides `setup.py`. `python -m pip install .` is alaready defined in the `abel_setup()` function.

.github/workflows/tools.yml
trav3711: `model_test.py` and `ontology_wrapper_test.py` rely on instance validator's simplified universe.

.github/workflows/tools.yml
trav3711: Done.

.github/workflows/tools.yml
trav3711: Done.

.github/workflows/tools.yml
trav3711: Done.

.github/workflows/tools.yml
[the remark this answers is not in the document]
trav3711: Done.

.github/workflows/tools.yml
trav3711: Done.

charbull (approved):

PR #781, opened 2022-09-23
Smoke test for instance validator presubmit workflow

trav3711:


charbull (commented):
Awseome ! it works :)
|well done

PR #785, opened 2022-09-23
Smoke test for type validator

trav3711:


trav3711:
Well...That's not good.

charbull (changes requested):
disable the type validator linter?

PR #813, opened 2022-10-26
Add default ontology path to silence ontology validator warnings

trav3711:
Silences ontology validator warnings AND speeds up instance validation since the default ontology is not being validated. A two-fer!

### DIFF
--- a/tools/abel/abel.py
+++ b/tools/abel/abel.py
@@ -81,8 +81,8 @@ def _spreadsheet_workflow(spreadsheet_id: str,
     print('Validating Export.')
     handler.RunValidation(
         filenames=[BC_EXPORT_PATH],
-        report_filename=INSTANCE_VALIDATOR_LOG_PATH,
-        modified_types_filepath=ontology_path,
+        report_filename=INSTANCE_VALIDATOR_LOG_PATH
+        default_types_filepath=ontology_path,
         subscription=subscription,
         service_account=service_account,
         timeout=timeout
@@ -110,7 +110,7 @@ def _bc_workflow(spreadsheet_id: str,
   handler.RunValidation(
       filenames=[bc_filepath],
       report_filename=INSTANCE_VALIDATOR_LOG_PATH,
-      modified_types_filepath=ontology_path
+      default_types_filepath=ontology_path
   )
   print(f'Importing Building Configuration file from {bc_filepath}.')
   imported_building_config = import_helper.DeserializeBuildingConfiguration(
--- a/tools/validators/instance_validator/validate/generate_universe.py
+++ b/tools/validators/instance_validator/validate/generate_universe.py
@@ -28,6 +28,7 @@ from yamlformat.validator import presubmit_validate_types_lib
 def BuildUniverse(
     use_simplified_universe: bool = False,
     modified_types_filepath: str = None,
+    default_types_filepath: str = constants.ONTOLOGY_ROOT,
 ) -> presubmit_validate_types_lib.ConfigUniverse:
   """Generates the ontology universe.
 
@@ -54,17 +55,17 @@ def BuildUniverse(
     external_file_lib.Validate(
         filter_text=None,
         changed_directory=modified_types_filepath,
-        original_directory=constants.ONTOLOGY_ROOT,
+        original_directory=default_types_filepath,
         interactive=False)
     yaml_files = external_file_lib.RecursiveDirWalk(modified_types_filepath)
   else:
-    default_ontology_exists = path.exists(constants.ONTOLOGY_ROOT)
+    default_ontology_exists = path.exists(default_types_filepath)
     if not default_ontology_exists:
       print(f'Specified filepath [{constants.ONTOLOGY_ROOT}] '
             'for default ontology does not exist')
       return None
     # use default location for ontology files
-    yaml_files = external_file_lib.RecursiveDirWalk(constants.ONTOLOGY_ROOT)
+    yaml_files = external_file_lib.RecursiveDirWalk(default_types_filepath)
 
   if yaml_files:
     config = presubmit_validate_types_lib.SeparateConfigFiles(yaml_files)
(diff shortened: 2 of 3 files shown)


tools/validators/instance_validator/validate/generate_universe.py
```diff
@@ -28,6 +28,7 @@
 def BuildUniverse(
     use_simplified_universe: bool = False,
     modified_types_filepath: str = None,
+    default_types_filepath: str = constants.ONTOLOGY_ROOT,
```
charbull, line 31: not sure I follow why we need to pass it here instead of reading it from the constants?

trav3711: because the constants file path won't work for applications that call the instance validator's Run Validation method outside of the instance validator.

For example, take ABEL which runs from `tools\abel\abel.py`. It will call `handler.RunValidation` from the instance validator which will then call `generate_universe.BuildUniverse`. If `BuildUniverse` uses  `constants.ONTOLOGY_ROOT` then this error will arise:

```
Specified filepath [/usr/local/lib/python3.10/dist-packages/validate/../../../../ontology/yaml/resources] for default ontology does not exist
```

This is because the `validate` package is being used but the constant is trying to reference the ontology as if constants exists in the digital buildings repo. `Setup.py` creates these packages. I originally had passed in a different a path from ABEL to the ontology as a modified types path, but then the ontology validator will run and print findings which is what we don't want. My final solution was just to add a default ontology argument so we can explicitly define the default ontology from any application that calls the instance validator instead of relying on the instance validator's constants.

Another solution would be to turn the ontology into a package, or somehow expose the ontology universe as a package with some kind of API.

tools/abel/abel.py
```diff
@@ -82,7 +82,7 @@ def _spreadsheet_workflow(spreadsheet_id: str,
     handler.RunValidation(
         filenames=[BC_EXPORT_PATH],
         report_filename=INSTANCE_VALIDATOR_LOG_PATH,
-        modified_types_filepath=ontology_path,
+        default_types_filepath=ontology_path,
```
charbull, line 85: this is very dirty hack to bypass the warnings.
What if the user didnt provide an ontology path here? aren't you deprecating the feature to allow users to run abel with a modified local ontology?

tools/validators/instance_validator/validate/generate_universe.py
charbull: why not update the constants.ONTOLOGY_ROOT?
here you are forcing the user to always set an ontology path?

charbull (changes requested):

tools/abel/abel.py
trav3711: Well the user never had the ability to pass in a local ontology. Previously I was passing in the DB ontology as a modified ontology which felt like a dirtier hack. 

Re your comment below: I'm not forcing the user to set the ontology path. I gave the default ontology path a default value of the path from the instance validtor. If a developer wants to use the instance validator's `RunValidation` method, now they have the option to pass in the DB ontology from a different location as well as a modified ontology. I'm almost positive that `constants.ONTOLOGY_ROOT` cannot be modified for sure outside of the instance validator.

charbull (changes requested):

tools/abel/abel.py
```diff
@@ -110,7 +113,8 @@ def _bc_workflow(spreadsheet_id: str,
   handler.RunValidation(
       filenames=[bc_filepath],
       report_filename=INSTANCE_VALIDATOR_LOG_PATH,
-      modified_types_filepath=ontology_path
+      modified_types_filepath=modified_types_filepath,
```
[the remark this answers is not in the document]
trav3711: sweet!

tools/validators/instance_validator/validate/handler.py
```diff
@@ -115,9 +116,11 @@ def RunValidation(filenames: List[str],
   try:
     print('\nStarting validator...\n')
     print('\nStarting universe generation...\n')
+    print(f'DEFAULT TYPES PATH: {default_types_filepath}')
     universe = generate_universe.BuildUniverse(
         use_simplified_universe=use_simplified_universe,
-        modified_types_filepath=modified_types_filepath)
+        modified_types_filepath=modified_types_filepath,
+        default_types_filepath=default_types_filepath)
```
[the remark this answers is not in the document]
trav3711: Added Args section.

tools/abel/abel.py
```diff
@@ -94,7 +96,8 @@ def _spreadsheet_workflow(spreadsheet_id: str,
 def _bc_workflow(spreadsheet_id: str,
                  bc_filepath: str,
                  gcp_token_path: str,
-                 ontology_path: Optional[str] = ONTOLOGY_ROOT) -> None:
+                 modified_types_filepath: Optional[str] = ONTOLOGY_ROOT
```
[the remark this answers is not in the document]
trav3711: Done.

tools/abel/abel.py
```diff
@@ -82,7 +83,8 @@ def _spreadsheet_workflow(spreadsheet_id: str,
     handler.RunValidation(
         filenames=[BC_EXPORT_PATH],
         report_filename=INSTANCE_VALIDATOR_LOG_PATH,
-        modified_types_filepath=ontology_path,
+        modified_types_filepath=modified_types_filepath,
+        default_types_filepath=ONTOLOGY_ROOT,
```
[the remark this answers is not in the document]
trav3711: Done.

tools/validators/instance_validator/validate/handler.py
```diff
@@ -100,12 +101,26 @@ def _ValidateTelemetry(subscription: str, service_account: str,
...
                   modified_types_filepath: str = None,
+                  default_types_filepath: str = None,
                   subscription: str = None,
                   service_account: str = None,
                   report_filename: str = None,
-                  timeout: int = 60,
+                  timeout: int = constants.DEFAULT_TIMEOUT,
                   is_udmi: bool = False) -> None:
-  """Master runner for all validations."""
+  """Master runner for all validations.
```
charbull, line 110: can you please change `master` to `coordinator` or something else

charbull (changes requested):

trav3711:
Also added in the new `modified-types-filepath` argument to the readme.

charbull (approved):

PR #844, opened 2022-11-22
[DX UNIT MODES ADDITION]

darshanjain12:


ontology/yaml/resources/states/states.yaml
```diff
@@ -27,6 +27,7 @@ VFD: Discrete speed (frequency) control
 BYPASS: Bypassing an element or entity.
 DEHUMIDIFY: Modes of operation in DX unit where the unit acts as a dehumidifier by removing moisture from air
 FAN_ONLY: Modes of operation in DX unit where only fan runs and heating/cooling is disabled
+SETPOINT: Mode of operation in DX unit where the unit runs as per the setpoint provided.
```
tasodorff, line 30: Doesnt it control to setpoint during the other modes?

darshanjain12: Hi Trevor
 The Setpoint mode is used when the unit is not the cool/heat master.

image

darshanjain12: Hi Trevor,
Update on the trailing comments .

 After the confirmation from the OEM found out that there is no configuration for master and slave here in this particular site.
I am removing the setpoint mode from the pull request as well.

tasodorff (approved):

PR #853, opened 2022-12-07
bug fix for gateways with missing entities.

tasodorff:
added logic and tests to handle missing fields on GATEWAYS/PASSTHROUGH entities.

### DIFF
--- a/tools/validators/instance_validator/tests/handler_test.py
+++ b/tools/validators/instance_validator/tests/handler_test.py
@@ -99,18 +99,6 @@ class HandlerTest(absltest.TestCase):
     with self.assertRaises(SyntaxError):
       helper.Validate(entities, config_mode, is_udmi=True)
 
-  def testMissingBuildingDoesntKillValidation(self):
-    """ Validation runs to completion even with missing building. """
-    try:
-      input_file = os.path.join(_TESTCASE_PATH, 'BAD','missing_building.yaml')
-
-      _RunValidation([input_file], use_simplified_universe=True)
-
-    except SyntaxError:
-      self.fail('ValidationHelper:Validate unexpectedly raised Exception '
-                'when building missing. This should be handled without '
-                'failing.')
-
   def testValidateMultipleInputFilesSuccess(self):
     try:
       input_file1 = os.path.join(_TESTCASE_PATH, 'GOOD', 'building_type.yaml')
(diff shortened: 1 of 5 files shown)


tools/validators/instance_validator/validate/handler.py
```diff
@@ -159,13 +159,6 @@ def RunValidation(filenames: List[str],
             'run telemetry validation. See here for more details: '
             'https://google.github.io/digitalbuildings/tools/validators/'
             'instance_validator/#telemetry-validation')
-  # pylint: disable=broad-except
-  # Using broad exception to catch any underlying error that might cause
-  # validation to fail
-  except Exception as ex:
```
rniemo-g, line 165: Why remove this broad catch block? without it, wouldn't the output be something like:

"<exception text>
[INFO] Instance validation completed."

which is confusing

tools/validators/instance_validator/validate/entity_instance.py
```diff
@@ -358,28 +358,40 @@ def _ValidateTranslation(self, entity: EntityInstance, is_udmi: bool = False
 
       # Check if the field is defined MISSING
       if isinstance(ft, ft_lib.UndefinedField):
-
-        # If the field is MISSING and REQUIRED, warn the user.
-        if not type_fields[qualified_field_name].optional:
+        # If the type is allows undefined fields, we won't know if the field
+        # is required or not explicitly. Throw a warning that the user
```
rniemo-g, line 362: nit: sentence doesn't end

tools/validators/instance_validator/validate/handler.py
tasodorff: it will output "instance validation complete", but it will also print the error stack. This info just shows it completed.

there is a separate message that tells the user whether validation succeeded: "[INFO]\tAll entities validated SUCCESSFULLY."

rniemo-g (approved):

tools/validators/instance_validator/validate/handler.py
trav3711: Best solution would be to open a bug to write more detailed exception handling for the handler to catch a wide variety of errors. For now, I'm okay with this change.

tools/validators/instance_validator/tests/handler_test.py
```diff
@@ -99,18 +99,6 @@ def testValidateOneBuildingExistFails(self):
...
       helper.Validate(entities, config_mode, is_udmi=True)
 
-  def testMissingBuildingDoesntKillValidation(self):
-    """ Validation runs to completion even with missing building. """
-    try:
-      input_file = os.path.join(_TESTCASE_PATH, 'BAD','missing_building.yaml')
-
-      _RunValidation([input_file], use_simplified_universe=True)
-
-    except SyntaxError:
```
trav3711, line 109: This is a weird test...glad we're removing this.

trav3711 (approved):

PR #855, opened 2022-12-12
Upgrade digitalbuildings tools to depend on Python 3.11

trav3711:
Upgrading python version for all tools to be >= Python 3.11


trav3711:
I can find the bug that keeps breaking the scoring tool. It's not any of the parsing logic because the scoring tool depends on the instance validator's `Deserilaize` method for that which passes its own tests. My hunch is something in python's standard library has changed causing the scoring tool to break with the 3.11 migration. 

I was thinking the `map()` method changed but no dice yet. There's also not much documentation for python 3.11 bugs yet, but the release is considered stable.

https://www.geeksforgeeks.org/python-map-function/

tasodorff:
@trav3711 is this still valid or should we close?

sypks:
patched scoring tool and running github actions now

sypks:
patch success. please review @trav3711 @tasodorff @charbull

tools/scoring/score/parse_config.py
```diff
@@ -181,8 +181,8 @@ def aggregate_translations(entity) -> List[Any]:
         return []
 
       translations[cloud_device_id] = {
-          f'{PROPOSED}': aggregate_translations(proposed_entity),
-          f'{SOLUTION}': aggregate_translations(solution_entity)
+          PROPOSED: aggregate_translations(proposed_entity),
```
trav3711, line 184: Why did the f string literal break the scoring tool? Is that a change in 3.11?

trav3711:
Looks like you need to get your CLA approved? Not sure how this is done but I thought you did this already?

tools/abel/setup.py
```diff
@@ -26,5 +26,5 @@
     install_requires=[
         'absl-py', 'pyfiglet', 'argparse', 'pyyaml', 'google-api-python-client'
     ],
-    python_requires='>=3.9',
+    python_requires='>=3.11',
```
charbull, line 29: I propose to keep it = for now so we dont encounter future changes/surprises, wdyt?

charbull (changes requested):
make the version stricter

PR #860, opened 2022-12-15
AHU-NEW-CANONICAL-TYPE-ADDITION

darshanjain12:
Adding AHU with chilled water based supply air temperature control, mixed air co2 based fresh air damper control, duct static pressure based vfd modulation


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -110,12 +110,16 @@ OAFC:
   opt_uses:
   - economizer_mode
   - mixed_air_temperature_sensor
+  - mixed_air_co2_concentration_setpoint
```
tasodorff, line 113: if its using this control instead of outside air flowrate & setpoint, i would create a new type. how do these fields interact with the OA flow control?

darshanjain12: I agree with you added separate abstract type for the same. 
The fields don't interact with OA flow control tried to imbibe in the existing abstract type.

charbull:
@tasodorff PTAL

PR #872, opened 2022-12-27
Test type guid generator workflow

jmact17:
This will run the ontology validator, which has been updated to attempt writing back to the ontology file


tasodorff:
@jmact17 is this still relevant? can we close this branch?

PR #880, opened 2023-01-05
added value ranges for telemetry fields

EugeniaDickson:


charbull:
hi @EugeniaDickson can you please take a look at the checks, there are some failures. Thank you

cstirdivant:
@charbull @jmact17 has the validator been updated to accept the new format for value ranges? I suspect this is the cause of a majority of the errors.

PR #889, opened 2023-01-15
UPDATE RO_CANONICAL TYPE

darshanjain12:


ontology/yaml/resources/PLUMBING/entity_types/RO.yaml
```diff
@@ -28,4 +28,7 @@ RO_STD:
   - control_mode
   - line_pressure_sensor
   - water_totaldissolvedsolids_concentration_sensor
+  opt_uses:
+  - high_pressure_status
```
cstirdivant, line 32: Can you please provide more detail about the intended use of this field?

cstirdivant (changes requested):

ontology/yaml/resources/PLUMBING/entity_types/RO.yaml
darshanjain12: This pressure status are used to turn On & Off the transfer pumps associated with the RO.

darshanjain12: Closing this pull request as low_pressure_status , high_pressure_status points are not been taken on cloud.

PR #896, opened 2023-01-25
Addition of a new VAV on supply side.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -262,6 +262,16 @@ CO2C:
...
   - zone_air_co2_concentration_setpoint
 
+CO2CDSP:
+  description: "Carbon dioxide control with dual setpoint."
+  is_abstract: true
+  implements:
+  - OPERATIONAL
+  uses:
+  - high_limit_zone_air_co2_concentration_setpoint
+  - low_limit_zone_air_co2_concentration_setpoint
```
tasodorff, line 272: What does it do if the CO2 is below setpoint? This point seems irrelevant, given the device cannot do anything if the value falls below the setpoint. 

Seems like the device would control like a normal `CO2C`-typed device. What am I missing?

tasodorff (commented):
Do not understand purpose of addition. Please address questions.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
ghairapetian: Hello,

The both setpoints are used for the regulation. If you go below the low_limit or above the high_limit, nothing special happens. However, in between, the damper's opening not the same. I attached a drawing to illustrate the regluation with 700 ppm as low_limit setpoint and 1100 ppm as high_limit setpoint.
IMG_20230130_135037

tasodorff (approved):

cstirdivant:
@ghairapetian please revert this PR to version [150a25f](https://github.com/google/digitalbuildings/pull/896/commits/150a25fa620eb6c01a371f740b7a9b8dfc6194e3)

charbull:
@ghairapetian friendly ping

cstirdivant (approved):

PR #907, opened 2023-02-04
SOLAR HOT WATER SYSTEM ADDITION

darshanjain12:


ontology/yaml/resources/PLUMBING/entity_types/ABSTRACT.yaml
```diff
@@ -34,4 +34,14 @@ LVLM:
   - level_status
   opt_uses:
   - percentage_sensor
+
+WTS:
```
tasodorff, line 38: this is called `WDT` in the HVAC namespace, so i would suggest using that name.  we can eventually elevate this concept to the global namespace and reference it in local namespaces, all under the same name.

tasodorff (changes requested):
Question on namespace appropriateness and request to adjust name of abstract type.

ontology/yaml/resources/PLUMBING/entity_types/ABSTRACT.yaml
darshanjain12: Yes. Committed the changes to add in the global name space

darshanjain12: Closing the pull request as we will be using HWS_SWTM canonical type.

PR #920, opened 2023-02-27
added high_hot_water_tank_temperature_alarm

kevin-hereworks:
added high_hot_water_tank_temperature_alarm


ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1773,4 +1773,7 @@ literals:
   - ACTIVE
   - INACTIVE
 
+- high_hot_water_tank_temperature_alarm:
```
tasodorff, line 1776: what entity will this be applied to? 

if you are planning on using this for a tank, the hot-water-tank bit is unnecessary; itll be obvious from context what the temperature alarm is for. the type would look something like this:
```
HWST_X:
  implements:
  - X
  - HWST
  uses:
  - high_water_temperature_alarm
```

kevin-hereworks: changed thank you, the entity has been added, Domestic hot water system with primary and secondary temperature monitoring

charbull:
@kevin-hereworks fyi the validator issue is fixed now, Thank you for your patience

tasodorff (approved):

PR #922, opened 2023-03-02
Dual section tank type

kevin-hereworks:
dual section water tank with temp, level monitoring

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -6487,3 +6487,16 @@ RFVSC2X:
   - return_fan_speed_percentage_command_2
   implements:
   - OPERATIONAL
+
+DSTLM:
+  description: "Dual section tank with level and temperature monitoring"
+  is_abstract: true
+  uses:
+  - section1_water_temperature_sensor
+  - section1_water_high_level_status
+  - section1_water_low_level_status
+  - section2_water_temperature_sensor
+  - section2_water_high_level_status
+  - section2_water_low_level_status
+  implements:
+  - OPERATIONAL
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -6499,4 +6499,4 @@ DSTLM:
   - section2_water_high_level_status
   - section2_water_low_level_status
   implements:
-  - OPERATIONAL
+  - MONITORING
--- a/ontology/yaml/resources/fields/telemetry_fields.yaml
+++ b/ontology/yaml/resources/fields/telemetry_fields.yaml
@@ -1773,4 +1773,20 @@ literals:
   - ACTIVE
   - INACTIVE
 
+# fields for dual section tank
+- section1_water_temperature_sensor
+- section2_water_temperature_sensor
+- section1_water_high_level_status:
+  - NORMAL
+  - HIGH
+- section2_water_high_level_status:
+  - NORMAL
+  - HIGH
+- section1_water_low_level_status:
+  - NORMAL
+  - LOW
+- section2_water_low_level_status:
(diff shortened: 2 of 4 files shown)


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6487,3 +6487,19 @@ RFVSC2X:
   - return_fan_speed_percentage_command_2
   implements:
   - OPERATIONAL
+
+DSTLM:
+  description: "Dual section tank with level and temperature monitoring"
+  is_abstract: true
+  uses:
+  - section1_water_temperature_sensor
```
tasodorff, line 6495: we generally use enumerations at the end for identical fields. you would instead have:
`water_temperature_sensor_1`
`water_temperature_sensor_2`
`water_high_level_status_1`
`water_high_level_status_2`
etc.

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -208,6 +208,8 @@ descriptor:
   source2: "Redundant power source 2 provision typically for Automatic transfer switch"
   scene: "The holistic ambience achieved from coordinated control of an area's lighting levels and colors. One of many stored setpoints that may be recalled to apply predetermined lighting levels across a number of fixtures."
   schedule: "Refers to the time-of-day run-time requirements for the equipment."
+  section1: "section 1 of a dual section tank"
```
tasodorff, line 211: Should not be necessary with enumeration.

tasodorff (changes requested):
Replace with enumerations.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
kevin-hereworks: removed thank you

charbull:
@kevin-hereworks any updates on this one? :)

kevin-hereworks:
Hi @charbull @tasodorff added optional alarm points and synced, this should be good to go

cstirdivant:
@tasodorff please review

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -3129,6 +3129,10 @@ literals:
     fixed_min: 0.0
     flexible_max: 2000.0
 
+- high_water_temperature_setpoint:
```
cstirdivant, line 3132: This is not descriptive enough; are you trying to model the threshold that will trigger an alarm or the control setpoint needed for the system?

cstirdivant (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
kevin-hereworks: yes, trying to model the threshold that will trigger the alarm

cstirdivant: Ok thanks for clarifying. Typically if you include the alarm point itself there is no need to model the threshold that triggers the alarm; I would suggest removing this field from your model.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6556,6 +6556,24 @@ RFVSC2X:
   implements:
   - OPERATIONAL
 
+DSTLM:
```
cstirdivant, line 6559: Suggest breaking this into two types (temp and level monitoring should be separated) re-naming to follow precedent. For example you can use the existing `LM` type and create a new one similar to it like `LM2X`. For temp monitoring, see `HWTTC` for an example.

cstirdivant (changes requested):

cstirdivant:
@kevin-hereworks how is this tank different from the ones you are modeling in https://github.com/google/digitalbuildings/pull/1025/files ?

PR #924, opened 2023-03-02
Addition of a new canonical CU type.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/CU.yaml
```diff
@@ -30,6 +30,18 @@ CU_SFSS_SFVSC_CHWSC_HTSC_SSPC_SARC_MTM:
   - SARC
   - MTM
 
+CU_DFSS_DFVSC_CHWZTC_HWZTC_RTM_ZHM:
+  description: "Discharge fan with start/stop, status, cooling and heating (zone temp control)."
```
tasodorff, line 34: You've set this as a compartment unit, not a discharge fan. This description seems wrong. What kind of equipment is this? 

Please note the definition for a [CU](https://github.com/google/digitalbuildings/blob/afcd129422941b8c4d50bcf8e268b251b1c700fb/ontology/yaml/resources/HVAC/entity_types/GENERALTYPES.yaml#L167) in our ontology. Does this match what your device does? What you've defined appears to be more of a `FCU` than anything.

tasodorff (changes requested):
Please confirm this matches the definition for a CU.

ontology/yaml/resources/HVAC/entity_types/CU.yaml
ghairapetian: I choose a CU because our equipement is served in air by an AHU. The FCU recirculates the air by taking air in the room and reinject it in the same room after cooling or heating.
This equipement don't.

charbull:
@tasodorff @ghairapetian any updates on this ?

ontology/yaml/resources/HVAC/entity_types/CU.yaml
cstirdivant: @ghairapetian  Trevor's comment is requesting that you revise the description for this type; it is not simply a "discharge fan" but rather a unit that performs heating and cooling. Can you please confirm that this unit is receiving outside air from an upstream plenum?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/CU.yaml
ghairapetian: Oh I am sorry. Didn't see the incorrect description. I confirm that this unit is receiving outside air from an upstream equipment (an AHU).

cstirdivant (approved):

PR #926, opened 2023-03-03
DSPZTC fan type added

kevin-hereworks:
dual setpoint monitoring and alarms for fan for mdf room

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -6487,3 +6487,17 @@ RFVSC2X:
   - return_fan_speed_percentage_command_2
   implements:
   - OPERATIONAL
+
+DSPZTC:
+  description: "Fan with dual setpoint control, high + low zone temperature alarms"
+  is_abstract: true
+  uses:
+  - run_status
+  - failed_discharge_fan_alarm
+  - zone_air_temperature_sensor
+  - high_limit_zone_air_temperature_setpoint
+  - low_limit_zone_air_temperature_setpoint
+  - high_zone_air_temperature_alarm
+  - low_zone_air_temperature_alarm
+  implements:
+  - MONITORING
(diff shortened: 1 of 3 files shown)


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6486,4 +6486,18 @@ RFVSC2X:
   - return_fan_speed_percentage_command_1
   - return_fan_speed_percentage_command_2
   implements:
-  - OPERATIONAL
\ No newline at end of file
+  - OPERATIONAL
+
+DSPZTC:
```
tasodorff, line 6491: Is there a reason you are using this over [DSP](https://github.com/google/digitalbuildings/blob/afcd129422941b8c4d50bcf8e268b251b1c700fb/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml#L256)?

high and low limit setpoints for zone air are basically what the heating and cooling setpoints on DSP do. Plus, some of the other stuff related to the discharge fan need to come out of this anyway since this functional control is really around zone temp control, not fan control.  In any case, the failed discharge fan and run status need to come out of this type. I would also recommend making the temperature alarms optional.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -736,4 +736,10 @@ FAN_SS_MSM:
   description: "Basic fan with speed monitoring (used in jet-fans for ventilation purpose)"
   implements:
   - FAN
-  - MSM
\ No newline at end of file
+  - MSM
+
+FAN_DSPZTC:
```
tasodorff, line 741: given the abstract type you put together, this seems to me like it is easily covered by the existing type [FAN_SS_DSP](https://github.com/google/digitalbuildings/blob/afcd129422941b8c4d50bcf8e268b251b1c700fb/ontology/yaml/resources/HVAC/entity_types/FAN.yaml#L67). it would give you all the same fields, assuming the cooling and heating setpoints are analagous to your zone temp limits.

tasodorff (changes requested):
Are zone temp limits the same as typical heating and cooling setpoints?

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
kevin-hereworks: Yes checked with people more knowledgeable than myself and this would work thank you.

PR #931, opened 2023-03-07
Update demo cloud device id

trav3711:
Updating demo.yaml cloud device id to comply with instance validator. Must be 16 digit string.


charbull (approved):

trav3711:
@charbull can you merge? I can't override the checks...

PR #932, opened 2023-03-07
Modify ontology validator action to generate type guids

jmact17:
Add automatic type guid generation for modified entity type ontology files


.github/workflows/OntologyValidator.yml
```diff
@@ -0,0 +1,77 @@
...
+        ref: ${{ github.event.pull_request.head.ref }}
+    
+    - name: Checkout head
+      if: ${{ github.event_name == 'push' }}
+      uses: actions/checkout@v3
+    
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
charbull, line 40: I recommend using another version as 3.7 end of support is in 3 months
https://www.python.org/downloads

see "Active Python Releases"

charbull (changes requested):

PR #956, opened 2023-03-12
New pmp type

kevin-hereworks:
Circulation Pump X2 with run command, run status and failed alarm for each


tasodorff (changes requested):
This is just two separate pumps. This should be modeled as two independent entities of type `PMP_SS`, which contains all the fields you want: `run_command`, `run_status`, and `failed_alarm`. Please let me know if there is an issue with that; otherwise, this extension should be unnecesary.

PR #960, opened 2023-03-14

cstirdivant:


cstirdivant:
@trav3711 please review the .md files and let me know which changes to keep and which to revert

ontology/docs/building_config.md
trav3711, line 1: I think the older version is the correct version on this one.

trav3711 (changes requested):
Revert `building_config.md`. Everything else looks good.

ontology/docs/building_config.md
jmact17: The newer version is more up to date I think, it has new value range and unit info. Is there merging that needs to be done?

jmact17 (approved):

trav3711 (approved):

PR #967, opened 2023-03-16
added failed alarm to optional field in co2c types

kevin-hereworks:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -298,6 +298,7 @@ CO2C:
   - zone_air_co2_concentration_sensor
   - zone_air_co2_concentration_setpoint
   opt_uses:
+  - failed_alarm
```
cstirdivant, line 301: This is not specific enough, what component has failed?

cstirdivant (changes requested):

PR #976, opened 2023-03-17
Update OntologyValidator action to only generate type guids on push/merge

jmact17:


ontology/yaml/resources/fields/telemetry_fields.yaml
jmact17, line 1: These are the same changes in PR #960, just needed them here to pass ontology validation

charbull (approved):

trav3711 (approved):

PR #989, opened 2023-03-28
Addition of only heating DOAS.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -79,4 +79,38 @@ DOAS_CHWSC_EFSS_EFVSC_FDPM3X_PHRU_HWSC_RCO2M_RSPC_SFM_SFSS_SFVSC_SSPC:
...
+
+DOAS_EFSS_EFVSC_FDPM3X_PHRU_HWSC_RCO2M_RSPC_SFM_SFSS_SFVSC_SSPC:
+  description: "Complex multi zone DOAS with heat recovery plate exchanger (heating only)."
+  is_canonical: true
+  implements:
+  - DOAS
+  - EFSS
+  - EFVSC
+  - FDPM3X
+  - PHRU
```
shambergoldstein, line 92: Can you clarify where the bypass section is for the system? Email screenshots to digitalbuildings-support@googlegroups.com.

shambergoldstein (dismissed):

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
shambergoldstein: Based on the provided screenshot, it appears that the bypass is for humidity control, yet none of these abstracts monitor/control humidity.

ghairapetian: Hello, I had the confirmation that this bypass is not for humidity control/monitoring. There is no sensor and no humidifier in this DOAS. The bypass is for the case we don't want to modify the outside air temperature regarding the temperature control. Is that ok for you ?

ghairapetian: friendly ping

ghairapetian:
I think I have answered to all requests. Do you need more information ?

PR #1001, opened 2023-04-03
new DWST canonical type and extend RMM abstract type

adelin-diac:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -4576,6 +4576,8 @@ RMM:
   guid: "162df361-7593-4e36-bf41-e7fa667244dd"
   description: "Run mode monitoring."
   is_abstract: true
+  opt_uses:
+  - run_status
```
shambergoldstein, line 4580: Use `RMM` with `SS` to bring in run_status.

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -140,6 +140,14 @@ PMP_SS_RMM:
   - low_pressure_status
   - high_pressure_status
 
+PMP_CPC_RMM:
```
shambergoldstein, line 143: Don't need to differentiate that it is a circulation pump, meaning you should be able to use `PMP_SS_RMM`

shambergoldstein (changes requested):

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -140,6 +140,14 @@ PMP_SS_RMM:
...
+  description: "Pump with circulation pump and run mode monitoring."
+    is_canonical: true
+    implements:
+    - PMP
+    - CPC
+    - RMM
```
cstirdivant, line 149: `RMM` is typically reserved for components that perform heating/cooling. It is odd to see this used on a pump.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
adelin-diac: the fields in CPC are also required for this entity. Is PMP_SS_RMM_CPC a valid name to add? I have updated to this for now.

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
adelin-diac: I have updated to PMP_SS_RMM_CPC for now. If its prefered to not use RMM, PMP_VSC_CPC could possible work if `run_command` and `speed_percentage_command` are marked missing, but unsure if using VSC and marking `speed_percentage_command` as missing would be valid.

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -140,6 +140,15 @@ PMP_SS_RMM:
...
   - high_pressure_status
 
+PMP_SS_RMM_CPC:
+  description: "Pump with circulation pump and run mode monitoring."
+  is_canonical: true
+  implements:
+  - PMP
+  - SS
+  - RMM
+  - CPC
```
cstirdivant, line 150: `CPC` should be used on non-PMP types. For example, if a boiler had a pump that was internal to the unit. Can you please clarify; are you trying to model a stand-alone pump or a circulation pump that is inside of another piece of equipment?

adelin-diac:
@cstirdivant apologies about delay with this device. There was some slight misunderstanding at the time of modelling this device. It is actually an air source heat pump, which serves domestic hot water. It has a circulation pump on it. Which type would be best to use in this case, PMP or create a general type for ASHP?

adelin-diac:
In terms of the `run_mode` states, the ones given to us are: 
- System Off
- System On
- Emergency Run
- Test Run

shambergoldstein:
@adelin-diac is this pr still needed?

adelin-diac:
Hi @shambergoldstein, I'm not sure. @kevin-hereworks would know if it is still needed.

PR #1005, opened 2023-04-05
Change ontology explorer to not require type guids

jmact17:


trav3711 (changes requested):
Tests seem to be failing.

PR #1009, opened 2023-04-13
Bug fix for udmi argument

trav3711:
The `--udmi` argument was never being set to false. the `default` and `action` arguments for Argparse were conflicting.

### DIFF
--- a/tools/validators/instance_validator/README.md
+++ b/tools/validators/instance_validator/README.md
@@ -102,8 +102,9 @@ Run `python instance_validator.py` and provide the following arguments:
 
   * `-t/--timeout` **[Optional]** The timeout duration in seconds for the telemetry validation test. The default value is 600 seconds, or 10 minutes. If this time limit is exceeded before the validator receives a test pubsub message for each of the entities configured in the given instance config file, the test will fail with an error and report the entities that were not heard from.
 
-  * `--udmi` **[Optional]** Validates entity metadata as [UDMI](https://github.com/faucetsdn/udmi/). Include argument when `True` and do not include when `False`.
-  `python instance_validator.py -i input.yaml --udmi` validates a building config against the udmi standard.
+  * `--udmi` **[Optional]** Validates entity metadata as [UDMI](https://github.com/faucetsdn/udmi/). Flag is set to `True` by default and include `--udmi=False` when not validating against udmi.
+  
+  `python instance_validator.py -i input.yaml` validates a building config against the udmi standard.
 
   * **NOTE:** The service account key and subscription are provided by the Google team. Please reach out to your IoT TPM for guidance.
 
--- a/tools/validators/instance_validator/README.md
+++ b/tools/validators/instance_validator/README.md
@@ -102,7 +102,8 @@ Run `python instance_validator.py` and provide the following arguments:
 
   * `-t/--timeout` **[Optional]** The timeout duration in seconds for the telemetry validation test. The default value is 600 seconds, or 10 minutes. If this time limit is exceeded before the validator receives a test pubsub message for each of the entities configured in the given instance config file, the test will fail with an error and report the entities that were not heard from.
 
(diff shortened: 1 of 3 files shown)


charbull:
@trav3711 Brian will start using the --udmi flag

tools/validators/instance_validator/instance_validator.py
```diff
@@ -92,14 +92,12 @@ def _ParseArgs() -> argparse.ArgumentParser:
...
       help='Absolute path to report output directory',
-      metavar='report-directory',
-  )
+      metavar='report-directory')
 
   parser.add_argument(
       '--udmi',
       dest='udmi',
       required=False,
-      default=True,
```
charbull, line 102: can we keep this please?

tools/validators/instance_validator/README.md
```diff
@@ -102,7 +102,8 @@ Run `python instance_validator.py` and provide the following arguments:
 
   * `-t/--timeout` **[Optional]** The timeout duration in seconds for the telemetry validation test. The default value is 600 seconds, or 10 minutes. If this time limit is exceeded before the validator receives a test pubsub message for each of the entities configured in the given instance config file, the test will fail with an error and report the entities that were not heard from.
 
-  * `--udmi` **[Optional]** Treat message stream on PubSub subscription as [UDMI](https://github.com/faucetsdn/udmi/). **NOTE:** This is required for telemetry validation when devices implement the UDMI specification. Set to true by default.
+  * `--udmi` **[Optional]** Validates entity metadata as [UDMI](https://github.com/faucetsdn/udmi/). Include argument when `True` and do not include when `False`.
```
charbull, line 105: keep the default for True?

trav3711:
@charbull This is a fix which came from Brian finding a bug with the `--udmi` flag.

charbull (changes requested):
Brian is already having error messages for the radix building config.
He will start using --udmi=False to pass the validation

tools/validators/instance_validator/README.md
trav3711: This was confusing for Brian. I think it's more straight forward if the flag is included, then the argument is true. Otherwise if the flag is not included the argument is false. 

Previously, the `default` and `action` arguments for `parser.add_argment` were conflicting and the `--udmi` flag was always true.

tools/validators/instance_validator/instance_validator.py
trav3711: See below.

tools/validators/instance_validator/README.md
charbull: since the msi are mainly using it for udmi, I prefer to keep the udmi mode set to true by default. wdyt?

charbull (changes requested):
wdyt?

tools/validators/instance_validator/README.md
trav3711: Done.

trav3711:
Set default = True.

tools/validators/instance_validator/instance_validator.py
```diff
@@ -92,28 +92,29 @@ def _ParseArgs() -> argparse.ArgumentParser:
...
+    is_udmi = False
   handler.RunValidation(
       filenames=args.filenames,
       modified_types_filepath=args.modified_types_filepath,
       subscription=args.subscription,
       service_account=args.service_account,
       report_directory=args.report_directory,
       timeout=int(args.timeout),
-      is_udmi=args.udmi,
+      is_udmi=bool(is_udmi),
```
charbull, line 119: I dont think you need the cast here anymore since you are setting the boolean values in your if statement

charbull (approved):

tools/validators/instance_validator/instance_validator.py
trav3711: Oh thanks for catching that. Done.

PR #1016, opened 2023-04-25
Addition of a new FCU type for Warm air curtain.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1689,3 +1689,9 @@ FCU_DSP_DFVSC_DTM_RTM:
   - DTM
   - RTM
 
+FCU_DFSMC_HWRC:
```
shambergoldstein, line 1692: `DFSMC` should have `DFSS` with it.

shambergoldstein (changes requested):

shambergoldstein (approved):

ontology/yaml/resources/states/states.yaml
```diff
@@ -18,9 +18,11 @@ AUTO: Running under automatic control.
 MANUAL: "Running under manual (hand) control."
 OPEN: Open position, typically for a valve or other pass-though.
 CLOSED: Closed position, typically for a valve or other pass-though.
+VERY_LOW : Very low speed or output setting.
```
cstirdivant, line 21: This is not an acceptable state because "VERY" is subjective. Can you please elaborate on the total number of discrete speed states this FCU has.

cstirdivant (changes requested):

ghairapetian:
Hello,
The speed states are OFF, 1, 2, 3, 4, 5
1 is the lowest, 5 is the maximum speed.
There are 6 total speed states including OFF.

cstirdivant:
> Hello, The speed states are OFF, 1, 2, 3, 4, 5 1 is the lowest, 5 is the maximum speed. There are 6 total speed states including OFF.

@ghairapetian 
Thank you for the info. Do you know if the speeds are linearly distributed? For example, could we safely assume that 1 = 20% speed, 2 = 20% speed, 3 = 40% speed, 4 = 60% speed, and 5 = 100% speed?

ghairapetian:
> > Hello, The speed states are OFF, 1, 2, 3, 4, 5 1 is the lowest, 5 is the maximum speed. There are 6 total speed states including OFF.
> 
> @ghairapetian Thank you for the info. Do you know if the speeds are linearly distributed? For example, could we safely assume that 1 = 20% speed, 2 = 40% speed, 3 = 60% speed, 4 = 80% speed, and 5 = 100% speed?

Hello,
No we can't. It could be distributed from 0 to 50%. Depends of the integration on site.

cstirdivant:
@ghairapetian thank you for the clarification - for your particular integration, we will need to know exactly what the speed distributions are for each piece of equipment with this type of fan speed control. What you will end up doing is adding the discrete numeric speed states to `discharge_fan_speed_mode` as follows (using the appropriate numbers, I am using 20, 40, 60, 80 as an example):
```
- discharge_fan_speed_mode:
  - AUTO 
  - LOW
  - MEDIUM
  - HIGH
  - OFF
  - VFD
  - 20_PERCENT ## in your example this would map to speed state 1
  - 40_PERCENT ## speed state 2
  - 60_PERCENT ## speed state 3
  - 80_PERCENT ## speed state 4, etc, etc
  ```

ghairapetian:
Correction done. I checked out and it seems to by linear up to 100% in my case.

However, seems to not pass the ontology validation script.

ghairapetian:
Done.

cstirdivant (approved):

PR #1024, opened 2023-05-10
new booster pump type

kevin-hereworks:


ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -3110,3 +3110,7 @@ literals:
 - ultraviolet_irradiance_sensor:
     fixed_min: 0.0
     flexible_max: 2000.0
+
+- boost_pump_run_status:
```
shambergoldstein, line 3114: Should just be able to use `run_status` for the booster pump.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6618,3 +6618,16 @@ VMADC:
   - mixed_air_damper_percentage_sensor
   implements:
   - CONTROL
+
+PSM2X:
```
shambergoldstein, line 6622: This seems to just be a combination of SS and WDPM for two booster pumps. See more in the comment below.

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -140,6 +140,14 @@ PMP_SS_RMM:
   - low_pressure_status
   - high_pressure_status
 
+PMP_PSM2X:
```
shambergoldstein, line 143: The `PMP` entity type should refer to a single pump rather than a collection. Based on my comment above, I would suggest having two `PMP` entities that are both `PMP_SS_VSC_WDPM`. Please check if that would work for your application.

shambergoldstein (changes requested):

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
kevin-hereworks: fixed, yes i can make this work

shambergoldstein (approved):

PR #1026, opened 2023-05-11
new control mode states

kevin-hereworks:


ontology/yaml/resources/fields/telemetry_fields.yaml
shambergoldstein, line 1: Can you explain what these control modes mean and perhaps provide screenshots of where you are seeing them in the BMS?

shambergoldstein:
The control modes are to indicate the signal to the device on whether it should be heating, cooling, etc. What temperature it controls to should be instead accounted for using abstract types such as `ZTC`, `STC`, etc. Are you able to use these for your device instead?

kevin-hereworks:
Okay I understand, yes i can use those abstract types

PR #1028, opened 2023-05-12
Hws type

kevin-hereworks:


ontology/yaml/resources/HVAC/entity_types/HWS.yaml
```diff
@@ -313,6 +313,17 @@ HWS_SS_HWTTC_PWDPM2X:
   opt_uses:
   - compressor_run_status
 
+HWS_CPVSC2X_SWTC:
```
shambergoldstein, line 316: Circulation pumps can be made their own `PMP` general types, perhaps `PMP_SS_VSC`, and `HWS_SWTC` can be used for the system.

shambergoldstein (changes requested):

ontology/yaml/resources/HVAC/entity_types/HWS.yaml
kevin-hereworks: okay, it it okay to have the 2 pumps modelled together or should they be split into 2 seperate pumps?

cstirdivant: are the circulation pumps on the system or integral to the boiler? if they are on the system they belong as their own PMP types. If they are integral to the boiler, they belong on the BLR types.

cstirdivant (changes requested):

PR #1030, opened 2023-05-12
New pmp psm type 

adelin-diac:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6618,3 +6618,14 @@ VMADC:
   - mixed_air_damper_percentage_sensor
   implements:
   - CONTROL
+
+PSM:
```
shambergoldstein, line 6622: Can `WDPM` be used here? If `differential_pressure_sensor` is missing then I would flag it as missing for your unit rather than creating a new type for the alarms/status only as the alarm logic likely depends on this sensor.

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -140,6 +140,13 @@ PMP_SS_RMM:
   - low_pressure_status
   - high_pressure_status
 
+PMP_PSM:
```
shambergoldstein, line 143: See comment on `PSM` above.

shambergoldstein (changes requested):

adelin-diac:
There is a pull request "ready to merge" that works for me now: #1024 that works for me. So closing.

PR #1031, opened 2023-05-12
unit heater with failed alarm

adelin-diac:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6618,3 +6618,11 @@ VMADC:
   - mixed_air_damper_percentage_sensor
   implements:
   - CONTROL
+
+FA:
```
shambergoldstein, line 6622: `failed_alarm` is included in `SS`, can this be used instead?

ontology/yaml/resources/HVAC/entity_types/UH.yaml
```diff
@@ -35,3 +35,10 @@ UH_SS_DSP_HTZC:
   - SS
   - DSP
   - HTZC
+
+UH_FA:
```
shambergoldstein, line 39: See above comment on `FA`.  Is `failed_alarm` truly the only point? What does the alarm logic consist of?

shambergoldstein (changes requested):

ontology/yaml/resources/HVAC/entity_types/UH.yaml
adelin-diac: Thanks for the comments. I will try to find out about alarm logic and get back to you.

PR #1033, opened 2023-05-15
New boiler and fan entities

annieDehghani:
Added a new boiler type. 
Added a new fan type.
Moved fields on to opt_uses  for generator, ATS and circ pump.

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/FAN.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/FAN.yaml
@@ -44,7 +44,7 @@ FAN_VSC:
   implements:
   - FAN
   - VSC
-
+  
 FAN_SS_RMM:
   guid: "917a9051-01dc-4f41-ad4c-5fcb45fe2f0e"
   description: "Fan with start/stop status and run mode monitoring"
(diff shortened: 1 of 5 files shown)


ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -36,7 +36,15 @@ FAN_SS_VSC:
   - FAN
   - SS
   - VSC
-
+  
+FAN_VSC:
```
tasodorff, line 40: This should not be needed, as `FAN_SS_VSC` exists. Please use that.

tasodorff (changes requested):
We cannot move fields from USES to OPT_USES. New types are required, but its probably more appropriate in your deployment to use MISSING fields in the BC translation, if they are truly missing.

tasodorff (changes requested):
We cannot move fields from USES to OPT_USES. New types are required, but its probably more appropriate in your deployment to use MISSING fields in the BC translation, if they are truly missing.

annieDehghani:
@tasodorff thank you for that feedback. I was not aware of the option to use MISSING in the translation. Thank you for pointing me to that.

PR #1036, opened 2023-05-16
Add spellchecker dependency and smoke test

trav3711:
Adding pyenchant dependency and smoke test with a typo.


tools/abel/model/model_builder.py
```diff
@@ -46,6 +46,8 @@
 class Model(object):
   """ABEL Concrete Model graph.
 
+  tyasdasdolll
```
charbull, line 49: 💯

charbull:
closing for now

PR #1041, opened 2023-05-19
Vibration sensor

kevin-hereworks:


shambergoldstein (changes requested):
Can you provide more details about the sensors/what type of vibration is being monitored (along what axes)? For example, see SENSOR_LV3A in the HVAC namespace.

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -363,6 +364,7 @@ measurement_descriptor:
   operating: "State when a decvice active"
   hue: "degree on color wheel from 0 to 360°"
   orientation: "Description of cardinal direction"
+  vibration: "Periodic back-and-forth motion."
```
cstirdivant, line 367: Please use `linearacceleration`  see example of existing vibration sensor here: `HVAC/SENSOR_LV3A`

ontology/yaml/resources/SAFETY/entity_types/ABSTRACT.yaml
```diff
@@ -94,3 +94,13 @@ LM:
   uses:
   - level_status 
 
+VA6X:
```
cstirdivant, line 97: You should not be mixing types of vibration being detected on the same type. These look like distinct types to me: one for the elevators, one for heaters, one for the gas_heaters. Furthermore, it seems like these are alarms that belong on existing types (e.g., the elevator alarms should be included on an existing `ELV` type, not modeled as a separate vibration sensor type.

cstirdivant (changes requested):

ontology/yaml/resources/SAFETY/entity_types/ABSTRACT.yaml
kevin-hereworks: okay, these can potentially be modelled as 3 different sensors `SENSOR_LV3A`, i can't add them to the devices at the moment

PR #1047, opened 2023-06-02
added abstract and canonical types

kevin-hereworks:


ontology/yaml/resources/SAFETY/entity_types/LDS.yaml
```diff
@@ -98,3 +98,12 @@ LDS_WLA_BCM:
   opt_uses:
   - water_leak_cable_distance_sensor
   - failed_alarm
+
+LDS_WLDS10X_WLS10X_WLA10X_BCM10X:
```
tasodorff, line 102: It seems like you have a bunch of stand-alone sensors that are communicating through one gateway device. I dont think we would want to model this 10x unit as a `canonical` thing, because I dont expect there is anything special about the fact that it happens to have 10 sensors. Perhaps an easier way to do this would be to define a simple canonical type, avoid any abstract types, and use virtual entities to define construct 10 individual leak sensors. 

If thats true, then you could define a new canonical type like this:
```
LDS_WLCC:
  description: Leak detection sensor with array of leak cable-based sensors."
  is_canonical: true
  uses:
  - water_leak_detection_alarm
  - water_leak_alarm
  - water_leak_cable_distance_sensor
  - broken_cable_alarm
```

Then, I would generate 10 virtual entities in your building config which point to the reporting device with those fields. You can use the type `GATEWAYS/PASSTHROUGH` for the reporting entity's type, and you won't need to define these 10x sensor abstract types.

tasodorff (changes requested):
Recommending that you simplify this device and break it out separately for each sensor. If you have any questions please let me know.

ontology/yaml/resources/SAFETY/entity_types/LDS.yaml
kevin-hereworks: this makes a lot more sense actually thank you

cstirdivant (approved):

PR #1056, opened 2023-06-12
fixed watt_per.. to watts_per..,added kilowatts_per

kevin-hereworks:


ontology/yaml/resources/units/units.yaml
```diff
@@ -234,7 +234,10 @@ level: distance
 impulse:
   newton_seconds: STANDARD
 irradiance:
```
shambergoldstein, line 236: Why add `irradiance` to the end of the unit?

shambergoldstein (changes requested):

ontology/yaml/resources/units/units.yaml
kevin-hereworks: the dbo doesn't allow using the same unit as powerflux (watts_per..) so I copied the format of percent_relative_humidity where humidity is at the end

PR #1068, opened 2023-06-19
new FAN_EFVSC type

adelin-diac:


ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -492,6 +492,13 @@ FAN_SS_VSC_SSPC_FDPSM:
   - VSC
   - SSPC
   - FDPSM
+
+FAN_EFVSC:
```
cstirdivant, line 496: Can you use the existing type `FAN_SS_VSC` to model what you need?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
adelin-diac: yes, that should work if changing the names of fields from `exhaust_fan_run_command` to `run_comand` etc. I was trying to model an exhaust fan with this entity type, so for future reference, when would the `exhaust_fan...` fields be used instead of the more general field names - would it be if the exhaust fan was part of a bigger unit being modelled?

cstirdivant: If the exhaust fan is part of a bigger unit, it should be modeled as part of the that unit (see examples like `AHU_BSPC_DFSS_DSP_DX2ZC_ECONZ_EFSS_EFVSC` where the exhaust fan functionality is captured by `EFSS` and `EFVSC`)

PR #1072, opened 2023-06-22
Spellcheck Ontology

azenla:


trav3711 (approved):
I like that the spellcheck isn't using setuptools. We should probably deprecate that package from the other tools as well.

trav3711 (changes requested):
Fix conflicts.

trav3711 (approved):
NICE!

Issue #1100, opened 2023-07-25
Ontology Explorer - Surface all field names from a list of subfields


cstirdivant:
I want to be able to pass in a list (with arbitrary delimiters) of subfields and want to see all the existing field names that use those subfields.

For example, if I pass in a list [supply, water, setpoint] the tool would return: supply_water_temperature_setpoint, chilled_supply_water_temperature_setpoint, heating_water_supply_temperature_setpoint, etc.

trav3711:
The above PR does add the desired feature. However some kind of error handling could be added in order to surface invalid subfields. Will add internally, and will keep issue open to track.

PR #1106, opened 2023-07-27
possible solution for issue #1100

adelin-diac:
we've been using a function internall that does exactly what #1100 requires.

Would appreciate any feedback on this!


trav3711:
Definitely a solution. Adding myself as a reviewer.

trav3711 (approved):

trav3711:
Testing before merging.

PR #1118, opened 2023-08-04
initial commit, discharge air temp added to ztc

kevin-hereworks:


ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1739,3 +1739,13 @@ FCU_DSP_DFVSC_DTM_RTM:
...
 
+FCU_DFVSC_RTM_ZTC_CO2M_HWVM_CHWVM:
+  description: "Fan coil unit with zone temperature control, heating water and chilled water valve monitoring."
+  implements:
+  - FCU
+  - DFVSC
+  - RTM
+  - ZTC
+  - CO2M
+  - HWVM
```
cstirdivant, line 1750: You should be using types like `HWZTC` and `CHWZTC`. That way you do not need to add discharge temp to `ZTC` (nor do you need to use `ZTC` at all).

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
kevin-hereworks: the reason i used ztc was because it had zone_air_deadband_temperature_setpoint, i have switched to hwztc, can i add zone_air_deadband_temperature_setpoint to this?

kevin-hereworks: i can also leave it off if you think there no need for it

cstirdivant (approved):

PR #1121, opened 2023-08-14
Addition of a new DOAS.

ghairapetian:
For information : the isolation dampers exist but I don't have any points for them. I don't know if I need to model them and mark the points as missing or remove the OIADC and RAIDC from the entity type submitted.


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6734,3 +6736,10 @@ OAQM:
   - outside_air_co2_concentration_sensor
   implements:
   - MONITORING
+
+OAIDC:
```
cstirdivant, line 6740: If you do not have visibility to these points, there is no need to model this at all.

ghairapetian: I removed them.

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -81,6 +81,22 @@ DOAS_CHWSC_EFSS_EFVSC_FDPM3X_PHRU_HWSC_RCO2M_RSPC_SFM_SFSS_SFVSC_SSPC:
...
+  is_canonical: true
+  implements:
+  - DOAS
+  - EFSS
+  - EFVSC
+  - SFSS
+  - SFVSC
+  - CHWSC
+  - HWSC
+  - RSPC
```
cstirdivant, line 95: I am assuming that the exhaust fan is controlling its speed to the `return_air_static_pressure_sensor`. If so, please use `EPC` instead of `RSPC`. Technically, since the air leaving the space can only be exhausted (since this is a DOAS unit, we really don't have "return_air"). I suggest re-naming all "return" fields to use "exhaust" instead.

cstirdivant (changes requested):

cstirdivant (approved):

PR #1122, opened 2023-08-14
Addition of a new FAN with 2 speed mode

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -769,3 +769,10 @@ FAN_SS_MSM:
   implements:
   - FAN
   - MSM
+
+FAN_HLC:
```
cstirdivant, line 773: Please include `SS` here as well to capture the general fan on/off status.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
ghairapetian: Done.

cstirdivant (approved):

PR #1129, opened 2023-08-18
Addition of a new CU with built-in dampers.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/CU.yaml
```diff
@@ -43,6 +43,21 @@ CU_DFSS_DFVSC_CHWZTC_HWZTC_RTM_ZHM:
...
+  implements:
+  - CU
+  - DFSS
+  - DFVSC
+  - CHWZTC
+  - HWZTC
+  - RTM
+  - ZHM
+  - CO2CDSP
+  - PDSCV
```
cstirdivant, line 58: Can you please clarify: is this damper upstream of the discharge fan?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/CU.yaml
ghairapetian: The supply damper is upstream indeed. The air from the DOAS go through the damper first then through the discharge fan.

cstirdivant: ok thank you for clarifying

cstirdivant (approved):

PR #1130, opened 2023-08-21
new energy meter with phase level energy monitoring

adelin-diac:


ontology/yaml/resources/METERS/entity_types/ABSTRACT.yaml
```diff
@@ -91,3 +91,10 @@ VCM:
   - current_sensor
   - voltage_sensor
   
+PLEM:
```
cstirdivant, line 94: This is rare to see -- is there known equipment in the building that is only connected to one of the three phases that is not independently sub-metered? (Typically equipment or panel level sub-meters are able to measure the "phase level energy" without having to measure it upstream like you are trying to do here.)

cstirdivant (changes requested):

PR #1131, opened 2023-08-21
ahu with heat  wheel

kevin-hereworks:


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -4100,6 +4100,23 @@ AHU_CHWSC_HWSC_ECON_EDBPC_RFSS2X_RFVSC2X_SFSS2X_SFVSC2X_SSPC:
...
   - SSPC
 
+AHU_SFVSC_EFVSC_HTWHLSTC_CHWVM_HWVM_SDM_EDM_SFM_FDPSM_CO2C_SDM2X_EDM2X:
+  description: "Multi-zone chilled water and heating water AHU with heat wheel, supply, exhaust fan speed control"
+  is_canonical: true
+  implements:
+  - AHU
+  - SFVSC
+  - EFVSC
+  - HTWHLSTC
```
cstirdivant, line 4110: If this a heat wheel that is transferring heat between the supply and exhaust streams, this equipment seems like a DOAS unit instead (where there is no mixed air and the supply/exhaust streams are isolated from each other).

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
kevin-hereworks: okay thank you, double checked and changed to DOAS

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -151,3 +151,21 @@ DOAS_CHWDC_EFSS_EFVSC_EFDPM_PHRUD_HWDC_DFSS_DFVSC_DSPM_DFDPM:
   - DSPM
   - DFDPM
 
+DOAS_SFVSC_EFVSC_HTWHLSTC_CHWVM_HWVM_SDM_EDM_SFM_FDPSM_CO2C_SDM2X_EDM2X:
+  description: "Multi-zone chilled water and heating water DOAS with heat wheel, supply, exhaust fan speed control"
+  is_canonical: true
+  implements:
+  - AHU
```
cstirdivant, line 158: You need to implement `DOAS` instead of `AHU`

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -151,3 +151,21 @@ DOAS_CHWDC_EFSS_EFVSC_EFDPM_PHRUD_HWDC_DFSS_DFVSC_DSPM_DFDPM:
...
+  - SFVSC
+  - EFVSC
+  - HTWHLSTC
+  - CHWVM
+  - HWVM
+  - SDM
```
cstirdivant, line 164: You do not need `SDM` and `EDM` if you are using the 2X types.

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -151,3 +151,21 @@ DOAS_CHWDC_EFSS_EFVSC_EFDPM_PHRUD_HWDC_DFSS_DFVSC_DSPM_DFDPM:
...
+  - EDM
+  - SFM
+  - FDPSM
+  - CO2C
```
cstirdivant, line 168: How can you have a single CO2 zone control with a multi-zone unit?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
kevin-hereworks: so this would be an average of the different co2 sensors, the control isn't different for the different zones

cstirdivant (changes requested):

cstirdivant (approved):

PR #1136, opened 2023-08-31
pump system controls 2 pumps

kevin-hereworks:


kevin-hereworks:
@cstirdivant this would be the fields coming from the pump unit, i will link out the individual fields `run_command
run_status_1
failed_alarm_1
speed_percentage_command_1` into the 2 individual pumps

cstirdivant:
Hi @kevin-hereworks there is no need to model the "controller" as a type, you can classify it as GATEWAY/PASSTHROUGH (the only entities that need legitimate types are the individual pumps)

cstirdivant:
@kevin-hereworks you can delete this PR (you will make the device sending all of this data type GATEWAYS/PASSTHROUGH). Then you will link out each field to its "virtual" device (e.g., HWP-1, HWP-2, HWS). The common fields should be attributed to the HWS.

PR #1137, opened 2023-08-31
added boiler type

kevin-hereworks:


ontology/yaml/resources/HVAC/entity_types/BLR.yaml
```diff
@@ -282,6 +282,14 @@ BLR_SS_RWTC_HTRWC:
   - RWTC
   - HTRWC
 
+BLR_HT2SWC_RWTM:
```
cstirdivant, line 285: Are there general `run_command` or `run_status` fields available? I would expect to also have the `SS` type here.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/BLR.yaml
kevin-hereworks: There is a run command (or heater_run_command) for the master boiler and the unit itself decides whether it needs the second boiler to run. there is a status field for both . I thought it would make more sense to model it with the heater_run_command_1 and the heater_run_status_1 heater_run_status_2

cstirdivant: So there are two distinct boilers? Each with one heater status? If so, each boiler would have a type more like `BLR_SS_HTSWC_RWTM` (where what you've called heater_run_status_1 would actually just be heater_run_status for the first boiler, and heater_run_status_2 is just heater_run_status for the second boiler).

kevin-hereworks: Apologies, so no its not 2 separate boilers, its 1 physical boiler with 2 modules/stages

cstirdivant (approved):

PR #1145, opened 2023-09-14
chiller with return chilled water temp control

kevin-hereworks:


kevin-hereworks:
checking the return vs supply control with bms experts at the moment, @cstirdivant does this look odd to you?

cstirdivant:
@kevin-hereworks it is uncommon to see return temp control, but it is not necessarily incorrect. This appears to be a water-cooled chiller (since you have condensing water fields), and I am surprised to not see other fields related to isolation valves, refrigeration circuits, or pressure.

kevin-hereworks:
@cstirdivant, there is a bypass valve on the DACs but not on the chillers, regarding the other fields we don't have any points that would relate to those fields

cstirdivant (approved):

PR #1150, opened 2023-09-19
Update handler.py

trav3711:
Interdependency check should raise a warning but not cause the Instance Validator to fail.


azenla (approved):
LGTM

PR #1151, opened 2023-09-21
Addition of a new MAU.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/MAU.yaml
```diff
@@ -668,6 +668,19 @@ MAU_DFSS_DFVSC_HWDDC_DX2DDC_ZTM_DFM:
...
+MAU_SFSS_SFVSC_HWSC_CHWSC_RSPC_FDPM2X:
+  description: "Makeup air unit with heating, cooling and return static pressure control."
+  is_canonical: true
+  implements:
+  - MAU
+  - SFSS
+  - SFVSC
+  - HWSC
+  - CHWSC
+  - RSPC
```
cstirdivant, line 680: By definition, MAUs do not have return air (they only provide conditioned outside air to the building).

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/MAU.yaml
ghairapetian: The correction is done. It was an error.

cstirdivant (approved):

cstirdivant:
@shambergoldstein please merge and add guids

PR #1154, opened 2023-09-22
Addition of a new AHU.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -918,6 +918,24 @@ AHU_DFSS_CHWDT_DTM_RTC_RWISOVPC_OAFC_FDPSM2X:
   - OAFC
   - FDPSM2X
 
+AHU_SFSS_SFVSC_EFSS_EFVSC_HTWHLZTC_HWZTC_CHWZTC_SSPC_RSPC_CO2DSPMADC_MTM_FDPM3X:
+  description: "Single zone AHU."
```
cstirdivant, line 922: You are calling this a single-zone unit but you are using supply fan abstract types, which indicates this is a multi-zone unit. Please rectify this.

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -3190,3 +3190,16 @@ literals:
   - ACTIVE
   - INACTIVE
 
+- mixed_air_damper_open_command:
```
cstirdivant, line 3193: Is the mixed air damper truly binary? Or can it modulate? Is there a single field that can represent the damper position (I.e., `mixed_air_damper_command` is OPEN/CLOSED)?

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -918,6 +918,24 @@ AHU_DFSS_CHWDT_DTM_RTC_RWISOVPC_OAFC_FDPSM2X:
   - OAFC
   - FDPSM2X
 
+AHU_SFSS_SFVSC_EFSS_EFVSC_HTWHLZTC_HWZTC_CHWZTC_SSPC_RSPC_CO2DSPMADC_MTM_FDPM3X:
```
cstirdivant, line 921: I would expect this to have some form of economizer control (ECONZ, perhaps). You have nothing to model the outside air damper right now.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
cstirdivant: Ok makes sense, please note this in the type description. Something like "AHU that serves multiple dampers all feeding a single zone."

cstirdivant:
@ghairapetian please address the outstanding comments.

1. No `outside_air_damper_percentage_command` field or ECON type is used (I would expect this)
2. Please confirm that the mixed air damper is binary (open / closed) only and *cannot* modulate to a percentage open

ghairapetian:
Hello @cstirdivant ,

1. I don't have any point for the outside air damper (either command or sensor). Is that mandatory ?
2. The mixed_air_damper is truly binary and it needs two points for controling it. 1 for opening and 1 for closing. A single data point can't represent the damper position.

cstirdivant:
@ghairapetian upon further discussion with my team, we believe this may actually be a `DOAS` type instead of an `AHU`.

- Can you please confirm if the unit is "single-pass" -- the supply and exhaust streams are completely isolated from each other (and thermal energy is transferred between the two flows via a heat wheel).
- Please provide more details about the mixed air damper - is it truly a damper that mixes return air with supply air? Or it is a bypass damper on the supply side that allows for outside air to bypass the heat wheel to provide extra fresh air to the building?).

If you have any screenshots from the BMS or drawings that would help clarify, please email to BOS-GPS@google.com

Additionally, depending on your deliverable timeline, if you would like to meet over a video call to discuss any of your PRs further please let us know at BOS-GPS@google.com.

Thank you!

ghairapetian:
Hi @cstirdivant , correction done. Thank you very much.

PR #1163, opened 2023-10-03
Revert abel readme

trav3711:
The ABEL README.md was overwritten and reverted to an earlier version. This PR reverts the overwrite and adds documentation on how to use the split functionality.


azenla (approved):
LGTM

PR #1189, opened 2023-11-21
new dry air cooler

kellyhw:
@kevin-hereworks


ontology/yaml/resources/HVAC/entity_types/DC.yaml
```diff
@@ -24,3 +24,13 @@ DC_SS_SWTC:
...
   - SWTC
+
+DC_SS_SWTC_CDWISOVM:
+  description: "Typical dry-cooler (closed loop sensible cooler)."
+  is_canonical: true
+  implements:
+  - DC
+  - SS
+  - SWTC
+  - CDWISOVM
```
cstirdivant, line 35: Instead of using `CDWISOVM` I suggest using a more-general isolation valve type (e.g., `SWISOVM` or `RWISOVM`)

The "condensing" description on the valve is superfluous since this is a dry cooler with only one type of water.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/DC.yaml
kellyhw: Hi Claire, we are waiting for confirmation regards the isolation vales position to return on that.

kellyhw: Claire, as suggested the isolation valve type has been changed to RWISOVM as it is a return valve according to contractor.

cstirdivant (approved):

PR #1198, opened 2023-11-28
Update pip install batch file

trav3711:
Updating pip_install.bat file to include scorer and explorer setup.


cstirdivant:
@trav3711 upon further inspection it appears the global_setup and pip_install files are running different commands (setup.py and pip install, respectively) -- it is important to have distinct files that co do this? Can we combine? Are the global_setup files for sure not needed anymore?

trav3711:
> @trav3711 upon further inspection it appears the global_setup and pip_install files are running different commands (setup.py and pip install, respectively) -- it is important to have distinct files that can do this? Can we combine? Are the global_setup files for sure not needed anymore?

`pip install .` and `py setup.py` are essentially the same command but one uses pip to install dependencies and the other uses setuptools which is being deprecated so I switched everything to pip. Thanks for catching that!

cstirdivant (approved):

tools/pip_install.bat
cstirdivant, line 1: Do we need to include sections for instance_validator and ontology_validator, as well (I am noticing the tools included in global_setup don't match all the tools included here)?

trav3711: If you expand the pip setup file, all of the tools are present.

cstirdivant (approved):

PR #1200, opened 2023-11-29
Addition of a new HX for a steam exchanger.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -7171,3 +7171,13 @@ FPHC:
   uses:
   - heating_water_valve_percentage_command
   - chilled_water_valve_percentage_command
+
+CWSWC:
```
cstirdivant, line 7175: I would instead make `condensing_water_...` = `steam_[supply/return]_...`

I understand why you would use condensing (since the steam probably does condense as it moves through the HX), but for this model we should use `steam` instead. Please specify if the steam isolation valve is on the supply or return side, as well. 

Do you have other points available to you? Like the steam temperature or pressure?

cstirdivant (changes requested):

cstirdivant:
@ghairapetian please address comments added to the individual files.

ghairapetian:
I wanted confirmation to be sure of what you expected because my english is not very good. I will consider I have it. I will try something.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -7171,3 +7171,13 @@ FPHC:
   uses:
   - heating_water_valve_percentage_command
   - chilled_water_valve_percentage_command
+
+SSTC:
+  description: Supply steam valve monitoring to control water temperature on supply side for steam exchanger.
```
cstirdivant, line 7176: Please add quotes around the description text.

ontology/yaml/resources/HVAC/entity_types/HX.yaml
```diff
@@ -187,3 +187,11 @@ HX_CWDT_HSWTC_HWDPC_HWBYPVPM:
   - HSWTC
   - HWDPC
   - HWBYPVPM
+
+HX_SSTC_SSSPC:
+  description: Steam exchanger with heating water temperature control and steam static pressure control on supply side.
```
cstirdivant, line 192: Please add quotes around the description text

cstirdivant (changes requested):

cstirdivant:
@ghairapetian the changes look good -- please add quotes to the description text.

PR #1204, opened 2023-12-01
New DHWT, Standard Calorifier with Return Pump

Ciaran-Oflannagain:
@cstirdivant - do you think the destratification pump is needed for this? Thanks


ontology/yaml/resources/HVAC/entity_types/DHWT.yaml
```diff
@@ -16,6 +16,16 @@
 ### Canonical Types  ###
 ########################
 
+DWHT_HTZTC_HWTTC_CPC_HWSWC:
+  description: "Standard calorifier with return pump."
+  is_canonical: true
+  implements:
+  - DHWT
+  - HTZTC
```
cstirdivant, line 24: This is the incorrect use for `HTZTC`; that type is typically used in air-side systems for an electric (or gas) heater that controls air temperature.

Ciaran-Oflannagain: Hi Claire - Apologies, the HTZTC was initially to cover the immersion electric heater (which is a backup and not the primary source source of heat) in this DHWT. I was unsure what abstract type to use, or if this point is needed in the cloud. I have removed the point. Thanks

cstirdivant: Hi @Ciaran-Oflannagain thanks for the clarification - if you do want to model the immersion heater, please use `HTSWC`

Ciaran-Oflannagain: Okay great thank you, I have added HTSWC.

Ciaran-Oflannagain:
@cstirdivant requested changes have been completed, thanks

cstirdivant (approved):

PR #1205, opened 2023-12-01
Addition of a simple Warm Air Curtain.

ghairapetian:


ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1338,6 +1338,16 @@ FCU_DFSS_CHWDC_CSP_RTM_DXZC:
   - filter_alarm
   - run_mode
 
+FCU_SS_RTC_STM_ZTM:
```
cstirdivant, line 1341: It is unclear how this equipment controls the return temperature -- it needs another type like DXRC, CHWRC, HTRC, HWRC, etc. (other examples in the ABSTRACT) to describe what component is used to control temperature.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1338,6 +1338,16 @@ FCU_DFSS_CHWDC_CSP_RTM_DXZC:
...
+  description: FCU with return temperature control, supply temperature monitoring, zone temperature monitorning and start/stop control
+  is_canonical: true
+  implements:
+  - FCU
+  - SS
```
cstirdivant, line 1346: The use of this is ambigious -- is it the SS control for the fan? Or for another component in the equipment? I would expect to see something like DFSS to indicate fan control.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
cstirdivant: Thank you for clarifying - upon further discussion with my team, we believe this is actually a `UH` type instead of a `FCU`. Can you confirm that the unit only runs if it is heating? (For example, the unit would never just run the fan, it will always run the fan + heater together)

If it always runs the fan + heater together, please reclassify as `UH_SS_HTRC_DTM_ZTM`
This would be a unit heater with start/stop, return temp control, discharge temp monitoring (changing from supply since the unit is providing air directly to the space) and zone temp monitoring.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
ghairapetian: Hello, 
This equipment can totally run the fan without heating if the temperature setpoint is reached.
I will try FCU_SS_HTRC_DTM_ZTM

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1338,6 +1338,16 @@ FCU_DFSS_CHWDC_CSP_RTM_DXZC:
   - filter_alarm
   - run_mode
 
+FCU_SS_HTRC_DTM_ZTM:
+  description: FCU with return temperature control, supply temperature monitoring, zone temperature monitorning and start/stop control
```
cstirdivant, line 1342: Please update the description (wrap with " ") and change supply --> discharge

cstirdivant (approved):

PR #1206, opened 2023-12-01
Addition of a new VAV.

ghairapetian:


cstirdivant (commented):
Please use existing `VAV_RD_CO2C` type instead (and use MISSING fields as necessary in your building config model)

PR #1212, opened 2023-12-05
Addition of a new FCU

Ciaran-Oflannagain:
AC unit with heating and cooling (zone temp control) @kevin-hereworks


Ciaran-Oflannagain:
@cstirdivant - Please see FCU PR (AC unit with heating and cooling (zone temp control)), when you have a moment. Thank you. Ciaran

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -19,6 +19,17 @@
 # This file defines fan-coil unit types. Valid types must include a fan and a coil of some type.
 # They must also include some control scheme (e.g. ZTC, DSP, etc.).
 
+FCU_SS_CO2M_ZTC_STC_RTM:
+  description: "AC unit with heating and cooling (zone temp control)"
+  is_canonical: true
+  implements:
+  - FCU
+  - SS
```
cstirdivant, line 27: This is ambiguous - is this SS control for the fan? The cooling/heating element?

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -19,6 +19,17 @@
...
+  - CO2M
+  - ZTC
+  - STC
```
cstirdivant, line 30: How can the unit control to both supply and zone temp?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
Ciaran-Oflannagain: Apologies, this was also just supply temp monitoring not control.

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
Ciaran-Oflannagain: Yes the SS is for the entire unit

cstirdivant: I would expect there to be separate SS for the fan (DFSS, etc) and the conditioning elements (DXZTC, HTZTC, etc)

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
Ciaran-Oflannagain: Okay thanks a million - we only have a single run command and status from the bms points but I will check them. Thank you

cstirdivant (changes requested):

PR #1235, opened 2023-12-20
update entity_instance.py and tests

trav3711:
Add validation for translations that have a base field name which is enumerated and un-enumerated in the same translation.


azenla (approved):
LGTM

PR #1244, opened 2024-01-10
Add exception handling to entity_instance.py

trav3711:
Add exception handling for the case when the reporting entity for a virtual entity either isn't present or doesn't have a translation.


azenla (approved):
LGTM

PR #1246, opened 2024-01-11
Remove html dropdown code

cstirdivant:


trav3711 (commented):
Why are you removing the html tags? Isn't that part of the markdown syntax?

cstirdivant:
Drop down is unnecessary -- was just testing to see how it affected readability and forgot to remove (drop downs not implemented in any other docs, so we want to be consistent)

trav3711 (approved):

PR #1248, opened 2024-01-12
Editing Workflow Method for getting Google Sheets API

Ciaran-Oflannagain:
@kevin-hereworks - How I got around the ValueError: Client secrets must be installed on web or app. Runs code, but then further errors later


trav3711:
@Ciaran-Oflannagain Would you like to add service account authentication to ABEL?

PR #1249, opened 2024-01-15
Variable speed pump with differential pressure and temperature monito…

Ciaran-Oflannagain:
…ring. @kevin-hereworks


ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -16,6 +16,15 @@
...
 ########################
 
+PMP_VSC_WDPC_WDT:
+  description: "Variable speed pump with differential pressure and temperature monitoring."
+  is_canonical: true
+  implements:
+  - PMP
+  - VSC
+  - WDPC
+  - WDT
```
cstirdivant, line 26: The pump is measuring temperature differential across itself? Is it a mixing pump?

Ciaran-Oflannagain: @cstirdivant - not mixing pumps. these are monitoring only for the temperature points, so not needed. I see there is an already approved type PMP_SS_VSC_WDPC in DBO, so I'll close this PR. Thank you

PR #1257, opened 2024-01-24
Add a new Gas Meter and Abstract Type

pranay2811:


pranay2811:
@cstirdivant Opened a new PR for Gas Meter. 
I see that a GUID was auto-generated for the abstract type, but it did not generate for the canonical type. Would you like me to manually add a new GUID for the canonical type?

tasodorff (approved):
LGTM

PR #1258, opened 2024-01-25
Added telemetry fields, Abstract Type and Canonical Type for FAN

pranay2811:
Added telemetry fields:
- zone_air_differential_pressure_sensor
- zone_air_differential_pressure_setpoint

Added Abstract Type: ZDPC

Added Canonical Type: FAN_SS_VSC_ZDPC

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -1676,6 +1676,7 @@ DSPM:
   - low_discharge_air_static_pressure_alarm
 
 ZDPC:
+  guid: feafa28d-9bee-4c4b-a506-5c372ac9d4eb
   description: "Zone differential pressure control."
   is_abstract: true
   uses:
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -1676,7 +1676,7 @@ DSPM:
   - low_discharge_air_static_pressure_alarm
 
 ZDPC:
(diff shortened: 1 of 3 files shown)


pranay2811:
@cstirdivant This is ready for review.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -1675,6 +1675,16 @@ DSPM:
   - high_discharge_air_static_pressure_alarm
   - low_discharge_air_static_pressure_alarm
 
+ZDPC:
```
tasodorff, line 1678: Explain. Differential pressure doesnt usually apply across a zone; thats typically static pressure. What specifically is being measured, and what is the differential across?

cstirdivant:
Hi @pranay2811 we have discussed internally and have concluded this model needs to be revised.

- Please use the existing fields `zone_air_static_pressure_sensor` and `zone_air_static_pressure_setpoint`
- Implement the existing abstract type `ZSPC` in your canonical type

We are choosing to use `static` instead of `differential` to follow the existing precedent set by `building_air_static_pressure_sensor` which is technically a differential between the building and the outside. Using the same logic, is is most-correct to use `zone_air_static_pressure_sensor` to represent the differential between an isolated zone and the rest of the building.

pranay2811:
Thanks for the clarification here @cstirdivant 
We will use the existing static-pressure fields as you mentioned. 
I will revise this PR to extend a new canonical type that implements ZSPC

pranay2811:
@cstirdivant This is ready for review. I have rolled back the previous commits & proposed a new canonical type for FAN with ZSPC

cstirdivant (approved):

PR #1273, opened 2024-02-16
Tools docs updates

cstirdivant:
Updating the original ontology path to point to the correct level of the directory

### DIFF
--- a/tools/validators/ontology_validator/README.md
+++ b/tools/validators/ontology_validator/README.md
@@ -45,11 +45,11 @@ The validator is Python-based and takes the following parameters:
 * `--modified-ontology-types` or `-m` **[Optional]**: An absolute or relative path to the modified files of the ontology.
 * `--interactive` or `-i` **[Optional]**: Enables interactive mode.
 
-The validator can be run as follows: `python3 validator.py -o=Users/foo/ontology/` or `python3 validator.py --original=Users/foo/ontology/`
+The validator can be run as follows: `python3 validator.py -o=Users/foo/ontology/yaml/resources` or `python3 validator.py --original=Users/foo/ontology/yaml/resources`
 
 ### Ontology Types Extended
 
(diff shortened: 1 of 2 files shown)


tools/validators/ontology_validator/README.md
```diff
@@ -45,11 +45,11 @@ The validator is Python-based and takes the following parameters:
 * `--modified-ontology-types` or `-m` **[Optional]**: An absolute or relative path to the modified files of the ontology.
 * `--interactive` or `-i` **[Optional]**: Enables interactive mode.
 
-The validator can be run as follows: `python3 validator.py -o=Users/foo/ontology/` or `python3 validator.py --original=Users/foo/ontology/`
+The validator can be run as follows: `python3 validator.py -o=Users/foo/ontology/yaml/resources` or `python3 validator.py --original=Users/foo/ontology/yaml/resources`
```
trav3711, line 48: I mean a user's ontology doesn't have to be under `ontology/yaml/resources`, that's just how we've done it. This still assumes that some poor soul out there has made their own ontology. For the sake of comprehension though, we could make this `path/to/original/ontology/`. Thoughts?

tools/validators/ontology_validator/README.md
```diff
@@ -45,11 +45,11 @@ The validator is Python-based and takes the following parameters:
...
 
 ### Ontology Types Extended
 
-If you have extended the ontology by adding new types to your local ontology, run the following: `python3 validator.py --original=Users/foo/ontology/ --modified-ontology-types=path/to/modified/ontology/types/folder`
+If you have extended the ontology by adding new types to your local ontology, run the following: `python3 validator.py --original=Users/foo/ontology/yaml/resources --modified-ontology-types=path/to/modified/ontology/types/folder`
```
trav3711, line 52: Same idea here as above.

trav3711 (changes requested):

tools/validators/ontology_validator/README.md
cstirdivant: Yes it does - we've had an MSI run into errors because there are folder trees in the `ontology` level that the validator is not happy about

trav3711 (approved):

PR #1285, opened 2024-02-29
Additional Fields and Entity Type for Lighting Fixtures

bsaxpi:
New Abstract "Color Temperature Control"
New Entity Type "LT_DC_CTC_OM"

### DIFF
--- a/ontology/yaml/resources/LIGHTING/entity_types/LT.yaml
+++ b/ontology/yaml/resources/LIGHTING/entity_types/LT.yaml
@@ -72,3 +72,13 @@ LT_DM_CHSLM_OM:
   - DM
   - CHSLM
   - OM
+
+LT_DC_CTC_OM:
+  guid: "f4127a88-ceab-46ed-bf74-0389b5699a00"
+  description: "Standard Lighting fixture with color temperature control and dimming monitoring."
+  is_canonical: true
+  implements:
+  - LT
+  - DC
+  - CTC
+  - OM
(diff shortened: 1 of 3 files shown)


ontology/yaml/resources/LIGHTING/entity_types/LT.yaml
```diff
@@ -72,3 +72,12 @@ LT_DM_CHSLM_OM:
   - DM
   - CHSLM
   - OM
+
+LT_DC_CTC_OM:
```
cstirdivant, line 76: Please add guid

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -3014,6 +3014,12 @@ literals:
 - lighting_color_hsl_hue_angle_setpoint:
     fixed_min: 0.0
     fixed_max: 360.0
+- lighting_color_temperature_setpoint:
+    flexible_min: 0.0
+    flexible_max: 10000.0
```
cstirdivant, line 3019: A max of 10,000 K is equivalent to 9726 degrees Celsius, which seems quite high

cstirdivant (changes requested):

ontology/yaml/resources/LIGHTING/entity_types/LT.yaml
bsaxpi: ...added

ontology/yaml/resources/fields/telemetry_fields.yaml
bsaxpi: Indeed. Changed to max. 7k. Should be sufficient.

cstirdivant (approved):

PR #1292, opened 2024-03-07
Added UPS battery remaining time

briancjohns:


ontology/yaml/resources/ELECTRICAL/entity_types/UPS.yaml
```diff
@@ -83,3 +83,17 @@ UPS_SS_UPSBM:
   - UPS
   - SS
   - UPSBM
+
+UPS_UPSBM:
```
cstirdivant, line 87: You cannot name this canonical type using an abstract type you aren't implementing. You need to fully implement `UPSBM` instead. We want to avoid defining individual fields on a canonical type.

Based on the set of fields you have here, it seems like you'd want to implement `UPSBM` and `IOBM`

cstirdivant (changes requested):

tools/tooling/Lib/site-packages/yaml/_yaml.cp312-win_amd64.pyd
cstirdivant, line 1: please remove these files from your PR

tools/tooling/Scripts/python.exe
cstirdivant, line 1: Please remove this file

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -2915,6 +2915,9 @@ literals:
   - LOW
   - MEDIUM
   - HIGH
+- remaining_battery_time_sensor:
```
cstirdivant, line 2918: use existing field `remaining_charge_time_sensor` instead

cstirdivant (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
briancjohns: Does "remaining_charge_time_sensor" indicate the amount of time remaining until the battery is fully charged, or the amount of time remaining until the battery is fully discharged?  I took it to mean the former, which leaves an important value out of the ontology, being the amount of time the UPS will continue to power the given load in battery mode.

tools/tooling/Lib/site-packages/yaml/_yaml.cp312-win_amd64.pyd
briancjohns: added tools/ to gitignore

tools/tooling/Scripts/python.exe
briancjohns: added tools/ to gitignore

ontology/yaml/resources/ELECTRICAL/entity_types/UPS.yaml
briancjohns: is_canonical: false

I don't want to implement IOBM because it contains quite a few required fields that we don't have.  UPSBM requires "remaining_charge_sensor", which we also don't have.  I've created a field called "remaining_battery_time_sensor" that accurately reflects the data we do have.

briancjohns: I've changed the type name to a non-canonical type name UPSTR

cstirdivant:
Hi @briancjohns I cannot review this until it only contains yaml files within the ontology/ folder

PR #1298, opened 2024-03-19
New CHWS and CDWS, abstracts, fields, subfields

evgeniya-engel:


evgeniya-engel:
@cstirdivant this one is ready for your review. Thank you!

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1269,6 +1287,9 @@ literals:
 - water_volume_accumulator:
     flexible_min: -0.000315451
     flexible_max: 0.315451
+- makeup_water_volume_accumulator:
+    flexible_min: -0.000315451
+    flexible_max: 0.315451
```
cstirdivant, line 1292: This is not a reasonable value range.

cstirdivant (changes requested):

ontology/yaml/resources/fields/telemetry_fields.yaml
evgeniya-engel: Would {min: 0, max: 10e6} be more adequate? Max value for water_volume_accumulator in our historical data is 6e6.

cstirdivant: what units? the units required here are cubic meters

evgeniya-engel: The reference data is in cubic meters. Actually 10e6 looks too large for cubic meters so I took 98th percentile instead and it's 95000 cubic meters. How about {min: 0, max: 1e5}?

cstirdivant (approved):

PR #1299, opened 2024-03-21
 Adding a New VRF

Ciaran-Oflannagain:
@kevin-hereworks


ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -19,6 +19,14 @@
 # This file defines fan-coil unit types. Valid types must include a fan and a coil of some type.
 # They must also include some control scheme (e.g. ZTC, DSP, etc.).
 
+FCU_DFMSC_DSP_ZTC:
```
cstirdivant, line 22: There are no abstract types here to capture the actual cooling control (e.g., DXDC, DXZC, etc.)

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -19,6 +19,14 @@
...
+  description: "Variable refrigerant flow unit with zone cooling and heating control, discharge fan speed modes"
```
cstirdivant, line 23: please add new GUID

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
cstirdivant: Hi @Ciaran-Oflannagain we encourage you to model the device as it exists in real life (not how it is digitally represented) -- we know that this has a DX control of some type (even if the fields are not available for integration). Based on what you already have, I would suggest adding DXZTC and marking the required fields as MISSING in your building config model.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
Ciaran-Oflannagain: @cstirdivant - thanks for your input I have added DXZTC and will make a note in building config. Thanks, Ciaran

shambergoldstein (approved):

PR #1301, opened 2024-03-26
New LGRP type

bsaxpi:
Lighting group type with two brightness percentage commands: One for the aisle faced luminaires and one for the window faced luminaires in an open space office


ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
```diff
@@ -30,6 +30,18 @@ DC:
   uses:
   - brightness_percentage_command
 
+LC_DUAL:
```
cstirdivant, line 33: while technically allowable, I suggest not creating an abstract type name like this to prevent confusion between abstract and canonical type names (this looks like a canonical type name)

I suggest naming this something like `AWIC` (Aisle window illuminance control)

ontology/yaml/resources/subfields/subfields.yaml
```diff
@@ -66,6 +66,7 @@ component:
 descriptor:
   air: "Atmospheric air, either conditioned or unconditioned."
   aisle: "Passage between two rows of server racks."
+  window: "Transparent part of the facade"
```
cstirdivant, line 69: Please place in alphabetical order within the descriptor section

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -1202,6 +1202,12 @@ literals:
 - dimmer_percentage_command: # Should be adjusted to brightness_percentage_command
     fixed_min: 0.0
     fixed_max: 100.0
+- aisle_brightness_percentage_command:
```
cstirdivant, line 1205: What does `aisle` represent in your use-case? It's typically used in server room applications, but I'm assuming it means something different here

cstirdivant (changes requested):

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
bsaxpi: Understood. AWIC sounds reasonable.

ontology/yaml/resources/subfields/subfields.yaml
bsaxpi: ...done

ontology/yaml/resources/fields/telemetry_fields.yaml
bsaxpi: These open offices usually have two rows of workplaces: One closer to the windows / facade and one row closer to the aisle or the door where one enters the office. So in this context aisle represents the hallway or corridor (thinking of it, "hallway* would maybe be the better subfield for this...)

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
```diff
@@ -30,6 +30,18 @@ DC:
   uses:
   - brightness_percentage_command
 
+LC_AWIC:
```
cstirdivant, line 33: Please remove the `LC_` so the type name is simply `AWIC` (the underscore is what makes this type name look confusing)

ontology/yaml/resources/fields/telemetry_fields.yaml
cstirdivant: Ok thank you for clarifying - it's ok to leave this as `aisle`

cstirdivant (changes requested):

ontology/yaml/resources/LIGHTING/entity_types/ABSTRACT.yaml
bsaxpi: ...done

cstirdivant (approved):

PR #1312, opened 2024-04-15
MAU with heating control and pressure status monitoring

Ciaran-Oflannagain:
Make up air unit with heating control using wet coil to maintain air flowrate and filter pressure status monitoring (2 sensors).


ontology/yaml/resources/HVAC/entity_types/MAU.yaml
```diff
@@ -16,6 +16,19 @@
 ### Canonical Types  ###
 ########################
 
+MAU_HWSC_SFC_SFVSC_SDM_FDPSM2X_SSPC:
+  guid: "042048f2-fde9-4330-a858-7d405e4996a0"
+  description: "Make-up air unit with heating control and filter pressure status monitoring"
```
cstirdivant, line 21: Is this a unit that serves downstream terminal units (I.e., VAVs)? If not, you want to use `discharge` instead of `supply`

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/MAU.yaml
Ciaran-Oflannagain: @cstirdivant - Definitely supply - It contains a supply fan, no extract fan and uses a heating wet coil to maintain air flowrate with two filter alarms. Thanks, Ciaran

cstirdivant (approved):

PR #1317, opened 2024-04-22
Adding new optional uses for HTWHLSTC

Ciaran-Oflannagain:
Cooling and heating percentage commands controlled by supply_air_setpoint


Ciaran-Oflannagain:
*temperature_setpoint

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -5861,6 +5861,8 @@ HTWHLSTC:
   - high_supply_air_temperature_alarm
   - low_return_air_temperature_alarm
   - low_supply_air_temperature_alarm
+  - heating_percentage_command
```
cstirdivant, line 5864: I'm not sure I understand how these are cooling/heating commands given to the heat wheel. The heat wheel can only modulate its rotational speed to transfer more or less energy between the supply and exhaust air streams. What do these fields mean in this context?

Ciaran-Oflannagain: Hi @cstirdivant - apologies for confusion. The heating and cooling commands are actually coils. These are a separate unit to the the heatwheel. The thermal wheel will heat supply air to a certain amount then the chilling/heating valves operate. I have changed the HTWHLSTC abstract type back and submitted the DOAS entity type. Thanks, Ciaran

Ciaran-Oflannagain:
Hi @cstirdivant  - please see updated DOAS entity type instead of new HTWHLSTC abstract type at your convenience. Thanks, Ciaran

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -16,6 +16,23 @@
...
+  - HTWHLSTC
+  - SFVSC
+  - EFVSC
+  - EDM
+  - SDM
+  - RFC
+  - SFC
+  - FDPM2X
+  - RTM
+  - FPHC
```
cstirdivant, line 34: FPHC is typically used for radiant heating/cooling (radiant beams, panels, etc.) -- why not use CHWSC and HWSC ?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
Ciaran-Oflannagain: Thank you for input  - have used both CHWSC and HWSC to cover heating and cooling valve percentage commands

cstirdivant (approved):

PR #1318, opened 2024-04-22
Adding a new AHU entity type and SATC Abstract Type

Ciaran-Oflannagain:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -1177,6 +1177,19 @@ SWTC:
   - supply_water_temperature_sensor
   - supply_water_temperature_setpoint
 
+# Rename SFN TOTAL_
+SATC:
```
cstirdivant, line 1181: This is generally discouraged; *how* does it control the supply temperature? With a heater? A DX unit? A chilled water coil?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
Ciaran-Oflannagain: @cstirdivant - sorry for delay, This AHU is provided a cooling_percentage command by the DX Unit. I'm struggling to add on the cooling_percentage_command. Can you give me any advice on adding it to the Entity Type. Thanks a million, Ciaran

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -17,6 +17,23 @@
...
+  - SFVSC
+  - EFVSC
+  - EDM
+  - SDM
+  - RFC
+  - SFC
+  - FDPM2X
+  - RTM
+  opt_uses:
+  - cooling_percentage_command
```
cstirdivant, line 36: We discourage using single fields on canonical types - what type of cooling element does this unit have? DX or CHW? Why not use CHWSC or DXSC?

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -17,6 +17,23 @@
 ########################
 
 ### Single Zone Units ###
+AHU_HWSC_SFVSC_EFVSC_EDM_SDM_RFC_SFC_FDPM2X_RTM:
```
cstirdivant, line 20: AHUs typically have some ECON control, you have no model for the mixed air or the outside air damper. This looks the same as your DOAS model. Please confirm what general type this should actually be.

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
Ciaran-Oflannagain: @cstirdivant - thanks it is a DX unit, I was avoiding using it as I didn't have points for the compressor run and status command. Have changed to DOAS, my mistake. Thank you

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
cstirdivant: Hi @Ciaran-Oflannagain I would use a full DX abstract type (e.g., DXSC) and mark whatever required points as MISSING (oif your model does not expose them).

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
cstirdivant: Thanks for the quick update

cstirdivant (approved):

PR #1322, opened 2024-04-23
Update TANK.yaml

cstirdivant:


shambergoldstein:
does this need to be added to the general types as well?

cstirdivant:
no, by adding the `/` before TK, it tells the ontology to reference the globally-defined TK type (so that we don't have to define TK within the HVAC folder)

shambergoldstein (approved):

PR #1327, opened 2024-05-02
Dual Tank Section

Ciaran-Oflannagain:
Dual Section Tank with Supply Water Temperature Monitoring and water level monitoring for both sections. @kevin-hereworks


ontology/yaml/resources/HVAC/entity_types/TANK.yaml
```diff
@@ -42,6 +42,15 @@ TK_LHLM_SWTM:
   - LHLM
   - SWTM
 
+TK_LHLM2X_SWTM2X:
+  guid: "e50038e9-60fc-4604-a632-5300043f93cc"
+  description: "Dual section Tank with supply water temperature monitoring with high and low water level monitoring."
```
cstirdivant, line 47: Can you please clarify what a dual section tank is? Are the two sections physically isolated, or can there be mixing between the two sections?

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/TANK.yaml
Ciaran-Oflannagain: Hi @cstirdivant  - apologies for the delay, have only received a response from the BMS engineers now. They are 1 Tank split with a divider but are joined by a common header. Thanks, Ciaran.

cstirdivant (approved):

PR #1328, opened 2024-05-02
New AHU Entity Type

Ciaran-Oflannagain:
AHU Entity type with Run around coil (RAC).


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -18,6 +18,24 @@
 
 ### Single Zone Units ###
 
+AHU_SS_HWSC_ETM_SFM_FDPSM_SFVSC_EFVSC_SDM_EDM_WDPM_VSC:
+  guid: "06e2e862-e9c2-46ec-91cd-d4b43cb44449"
+  description: "Single zone AHU with supply and return fans, with supply heating from run around coil and differential pressure monitoring."
+  is_canonical: true
+  implements:
+  - AHU
+  - SS
```
cstirdivant, line 27: This is discouraged and redundant if you have SFVSC

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -18,6 +18,24 @@
 
 ### Single Zone Units ###
 
+AHU_SS_HWSC_ETM_SFM_FDPSM_SFVSC_EFVSC_SDM_EDM_WDPM_VSC:
```
cstirdivant, line 21: Again, your AHU does not have any ECON modeled and is therefore incomplete. Please confirm if this is an AHU or DOAS

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -18,6 +18,24 @@
...
+  - SS
+  - HWSC
+  - ETM
+  - SFM
+  - FDPSM
+  - SFVSC
+  - EFVSC
+  - SDM
+  - EDM
+  - WDPM
```
cstirdivant, line 36: Differential pressure across what? There's a bunch of components in this equipment that this could apply to.

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -18,6 +18,24 @@
...
+  - VSC
```
cstirdivant, line 37: Speed control of what? the fan? this is not descriptive enough

cstirdivant (changes requested):

Ciaran-Oflannagain:
@cstirdivant - apologies I hadn't correctly modeled the circulation pump across the Run around coil connected to AHU. This is what VSC and WDPM are measured across. In terms of AHU vs DOAS. It's my understanding from commissioning engineers that the Economiser mode from the AHU is fixed by the commissioners to a comfort setpoint, so points are not needed regarding this as it will not be put in economy mode but it is still and AHU . Thanks, Ciaran

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
cstirdivant: If this is for the circulation pump, remove from here and model the pump separately

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -18,6 +18,24 @@
 
 ### Single Zone Units ###
 
+AHU_HWSC_ETM_SFM_FDPSM_SFVSC_EFVSC_SDM_EDM_CPC_WDPM_VSC:
```
cstirdivant, line 21: @Ciaran-Oflannagain even if the Economizer is fixed, we still must model it here -- DBO attempts to model things as they operate in real life, not as they digitally represent. I would imagine you still have data for a mixed_air_temperature_sensor or and outside_air_damper_percentage_command. Please use ECON and make necessary fields MISSING in your building config

ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -18,6 +18,22 @@
 
 ### Single Zone Units ###
 
+AHU_ECON_HWSC_ETM_SFM_FDPSM_SFVSC_EFVSC_SDM_EDM:
+  guid: "06e2e862-e9c2-46ec-91cd-d4b43cb44449"
+  description: "Single zone AHU with supply and return fans, with supply heating from run around coil."
```
cstirdivant, line 23: @Ciaran-Oflannagain just noticed this, if this is a single zone unit you need to use `discharge` instead of `supply` and use the appropriate abstracts (ECOND, HWDC, etc.)

cstirdivant (changes requested):

cstirdivant (approved):

PR #1330, opened 2024-05-02
elevate PMP and VLV to the global namespace

cstirdivant:


ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -154,30 +154,7 @@ PMP_SS_VSC_HWBYPVM:
...
+  - /PMP # inherits from global namespace
   - SS
   - VSC
   - HWBYPVM
-
-###################################
-### Existing Non-standard Types ###
-###################################
-
-PMP_SS_VSC_NONCANONICAL_1:
```
shambergoldstein, line 166: This type is used by two real world devices

PR #1341, opened 2024-05-15
Meter load connections

shambergoldstein:


ontology/docs/meter_loads.md
```diff
@@ -0,0 +1,91 @@
+# Meter Load Designation
+
+Digital Buidlings provides abstract models that can be used to represent various
```
cstirdivant, line 3: typo in buildings

ontology/docs/meter_loads.md
```diff
@@ -0,0 +1,91 @@
...
+to indicate that the load exists within the building.
+
+Example loadtype entity:
+``` yaml
+# The abstract entity for MAIN load
+abc-123:
+  code: MAIN
+  type: METERS/LOADTYPE_MAIN
+  connections:
+    ccb-342: CONTAINS
```
cstirdivant, line 61: Add building entity here just for clarity

ontology/docs/meter_loads.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Connect the Meter Entity to the Loadtype Entity via a `MEASURES` Connection
+
+Add a `MEASURES` connection to the meter entity to indicate which loadtype 
+the meter provides data for. A building may have multiple meters that 
+measure the same loadtype.
+
+Example `MEASURES` connection:
+``` yaml  
+# The updated reporting entity for the meter
```
cstirdivant, line 72: Either add all entities here (building, meter, loadtype) to show the complete model or create a new section for the full example model

ontology/yaml/resources/connections/connections.yaml
```diff
@@ -22,3 +22,5 @@ HAS_PART:
   description: "Source has some component or part defined by Target."
 HAS_RANGE:
   description: "Source has a coverage or detection range defined by Target."
+MEASURES:
+  description: "Source provides data that is representative of Target load."
```
cstirdivant, line 26: This is not a good enough description; need to revise to remove the specificity about "load" -- measures is a more abstract concept where it defines the relationship between a source quantifying an attribute defined by the target

tasodorff (approved):
LGTM, although would prefer connection between building and loadtypes to be something other than `contains` because that connection type doesn't really apply to abstract concepts. Its for physical connections. Can change in future.

PR #1344, opened 2024-05-20
Updated BMS for Fan

Ciaran-Oflannagain:
@cstirdivant Updated BMS points received which remove SS for failed_alarm and run_status. Thanks, Ciaran


cstirdivant (commented):
Hi @Ciaran-Oflannagain no need to remove `SS` since `run_status` and `run_command` are also required on `VSC` -- this modification is uncessary

PR #1347, opened 2024-05-27
New FCU Type

bsaxpi:
Added FCU Type FCU_DFMSC_ZTC - Fan coil unit with zone temperature control and discharge fan multi-speed control.


ontology/yaml/resources/HVAC/entity_types/FCU.yaml
```diff
@@ -1973,3 +1973,11 @@ FCU_DFVSC_RTM_CO2M_HWZTC_CHWZTC:
   - CO2M
   - HWZTC
   - CHWZTC
+
+FCU_DFMSC_ZTC:
```
cstirdivant, line 1977: How does the FCU control zone temperature? Does it have chilled water? A DX coil? Even if none of the points are digitally available, please include some form of conditioning/temperature control (replace ZTC with CHWZTC/DXZTC/etc)

cstirdivant (changes requested):

ontology/yaml/resources/HVAC/entity_types/FCU.yaml
bsaxpi: This is the issue I'm struggling with... The device is a Kampman Venkon (a water-cooled FCU) which definitely has a chilled water valve, but after months of discussions the BMS contractor is not able to make this field available. Basically it would be CHWZTC but then I had to mark the field "chilled_water_valve_percentage_command" as MISSING. Any ideas?

cstirdivant: Yes please add CHWZTC to this and mark the required fields as MISSING in your configuration. Thanks

bsaxpi: ...done

cstirdivant (approved):

PR #1351, opened 2024-05-28
Migrating shade_orientation_specification to shade_orientation_mode

ohshazbot:
Specification is an illegal point type as they made only have numeric values. Should be something that supports a Multi-state, like mode

Setting up in a standalone PR as this is potentially disruptive due to lack of backwards compatability


shambergoldstein:
field has been changed to shade_orientation_command; closing

PR #1353, opened 2024-05-31
Exhaust Fan with VSC and Zone Temperature Monitoring

Ciaran-Oflannagain:


ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -405,6 +405,16 @@ FAN_SS_EFVSC:
...
   - EFVSC
 
+FAN_SS_EFVSC_ZTM:
+  guid: "fd195953-7123-4414-ae9f-bb51bbc1eda8"
+  description: "Exhaust fan with variable speed control and zone temperature monitoring."
+  is_canonical: true
+  implements:
+  - FAN
+  - SS
+  - EFVSC
```
cstirdivant, line 415: This is the incorrect use of `EFVSC` (no need to specify exhaust if you're already modeling an exhaust fan). Why can't you use the existing `FAN_SS_VSC_ZTM' type?

Ciaran-Oflannagain:
@cstirdivant - thanks for all your help, that works perfectly. apologies in delay closing PR.

PR #1386, opened 2024-08-06
Add a new abstract to a new AHU type

DaoChen701:


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -5428,6 +5428,37 @@ AHU_STDSPC_SSPC_EFFC_EFVSC_FDPM3X_HTWHLSTC_RHM_ETM:
...
+  - SFSS2X
+  - SFVSC2X
+  - EFSS2X
+  - EFVSC2X
+  - EDBPC
+  - DX4SC
+  - ECON
+  - SSPC
+  - SARC
+  - HTWHLSS
```
cstirdivant, line 5460: How does an AHU have heat recovery? This seems more like a DOAS to me... please fix.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -6072,6 +6080,16 @@ HTWHLSTC:
   - supply_air_temperature_sensor
   - supply_air_temperature_setpoint
 
+HTWHLSS:
```
cstirdivant, line 6083: typically, heat wheels control to something (temperature, enthalpy, etc.) this is not a complete enough type. please fix.

cstirdivant (changes requested):

DaoChen701:
Hi Claire,

Thank you for the respond. This is an energy wheel between the return air duct and supply air duct. We are also in contact with the BMS team to clarify how to define this wheel. If you available, would you like have a short meet? I would like to introduce myself. Thank you!

Best,
Dao
________________________________

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
```diff
@@ -219,3 +219,18 @@ DOAS_SFVSC_EFVSC_HTWHLSTC_CHWVM_HWVM_SFM_FDPSM_ACO2C_SDM2X_EDM2X:
...
+  description: "DOAS with dual supply and exhaust fans, four stage compressor, building pressure control and energy recovery wheel"
+  implements:
+  - DOAS
+  - SFSS2X
+  - SFVSC2X
+  - EFSS2X
+  - EFVSC2X
+  - EDBPC
+  - DX4SC
+  - ECON
```
ccquigley, line 233: Since this is a DOAS unit, there won't be a mixed air temperature since it is direct outside air. The BMS lists this as mixed air which is a bit confusing. After discussing with Claire please create a new field, leaving_heat_recovery_wheel_temperature_sensor, add this to HTWHLSTC as opt uses and remove econ.

ccquigley (changes requested):
One change requested for the DOAS, other changes look good.

ontology/yaml/resources/HVAC/entity_types/DOAS.yaml
ccquigley: Edit: Please move this to the canonical section and add the canonical boolean

ccquigley (changes requested):

ccquigley (approved):

PR #1387, opened 2024-08-08
added failed_comms_alarm to HM

kevin-hereworks:


ontology/yaml/resources/METERS/entity_types/HM.yaml
```diff
@@ -26,6 +26,7 @@ HM_STANDARD:
   - thermal_energy_accumulator
   opt_uses:
   - water_volume_accumulator
+  - failed_communication_alarm
```
ccquigley, line 29: Please remove the failed_communication_alarm from this abstract and add it the the meter general type for HM. This way it will match the president set in HVAC.

ccquigley (changes requested):

ontology/yaml/resources/METERS/entity_types/HM.yaml
kevin-hereworks: done thank you

ccquigley (approved):

PR #1388, opened 2024-08-08
added failed_coms_alarm to SS

kevin-hereworks:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -754,6 +754,7 @@ SS:
   - failed_alarm
   - local_override_alarm
   - lost_power_alarm
+  - failed_communication_alarm
```
ccquigley, line 757: This alarm should be added to the equipment general type. Check the general type yamls file to see if it is already included in the equipment for SS. If not, please add it there instead of in SS.

kevin-hereworks: apologies, this is already on the general type. thank you

PR #1389, opened 2024-08-15
Adding Damper Canonical type

darshanjain12:
Adding Return Air Isolation Damper system: DMP_RAIDC

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/DMP.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/DMP.yaml
@@ -99,6 +99,14 @@ DMP_ZSPC_EFM:
   - ZSPC
   - EFM
 
+DMP_RAIDC:
+  guid: "6e70b2b3-542c-458e-8169-9820919853b2"
+  description: "Controlling and monitoring return air isolation damper system."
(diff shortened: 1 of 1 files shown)


ontology/yaml/resources/HVAC/entity_types/DMP.yaml
```diff
@@ -99,6 +99,14 @@ DMP_ZSPC_EFM:
   - ZSPC
   - EFM
 
+DMP_RAIDC:
```
ccquigley, line 102: Is this return air damper part of a larger system are there more points available? When possible we want to model the entire system instead of one part. For example, if this is a return air damper for a air handling unit, then we want to model the entire air handling unit with an abstract that includes the return air damper points.

ccquigley (changes requested):

ontology/yaml/resources/HVAC/entity_types/DMP.yaml
darshanjain12: Alright, I'll submit a pull request for the FCU canonical type. Thank you; I'm closing this pull request now.

PR #1398, opened 2024-08-22
HWS with Supply Temp Contol and valve monitoring

Ciaran-Oflannagain:


shambergoldstein (commented):
Please confirm what the electric heating is referring to. Is this a boiler or domestic hot water tank?

Ciaran-Oflannagain:
@shambergoldstein - update from BMS engineers is there isn't valve monitoring so just the supply temp control, which is a prexisting entity type. Closing PR, thanks for your assistance

PR #1399, opened 2024-08-22
Dual Pump Entity Type Hereworks

Ciaran-Oflannagain:
@kevin-hereworks

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/PMP.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/PMP.yaml
@@ -60,7 +60,7 @@ PMP_SS_VSC_WDPM:
   - VSC
   - WDPM
 
-PMP_SWTM_CPVSC2X:
+PMP_CPVSC2X_SWTM:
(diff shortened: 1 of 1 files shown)


ontology/yaml/resources/HVAC/entity_types/PMP.yaml
```diff
@@ -60,6 +60,15 @@ PMP_SS_VSC_WDPM:
   - VSC
   - WDPM
 
+PMP_CPVSC2X_SWTM:
```
shambergoldstein, line 63: The pump general type is intended to representative of a single pump rather than a set of pumps. Please model these pumps seperately and use a passthrough to contribute any shared points to both entities.

shambergoldstein (changes requested):

ontology/yaml/resources/HVAC/entity_types/PMP.yaml
Ciaran-Oflannagain: @shambergoldstein - will do thanks a million for your assistance.

PR #1403, opened 2024-08-26
Updating MAU Entity Type

Ciaran-Oflannagain:
@kevin-hereworks - AHU-18301.AHU-18302


Ciaran-Oflannagain:
@ccquigley - Updating previously submitted MAU entity type to remove SSPC and FDPM, changing SFM to SFC from monitoring to control.

ccquigley (approved):
Nothing has been on boarded with this type and updates to implements and name look correct.

PR #1429, opened 2024-09-24
Chiller abstract changes

DaoChen701:
Change a new chiller type.


shambergoldstein (approved):

shambergoldstein:
Note: Not a breaking change - editing type that was just pushed

PR #1433, opened 2024-09-25
Removed extra spacing under heating_run_status

pranay2811:


pranay2811:
@shambergoldstein I noticed that there was extra spacing before the heating_run_status values. I have removed them. Let me know if this is the correct format.

shambergoldstein (approved):

shambergoldstein:
good catch, thanks!

PR #1450, opened 2024-10-24
hydronic system additions / edits

evgeniya-engel:


evgeniya-engel:
@shambergoldstein @ccquigley ready for your review! I verified that the edited types are only used in only these assets that they were created fore and no other assets will be affected.

ontology/yaml/resources/units/units.yaml
```diff
@@ -102,6 +102,8 @@ distance:
   inches:
     multiplier: 0.0254
     offset: 0
+thermalefficiency:
+  kilowatts_per_ton: STANDARD
```
shambergoldstein, line 106: The telemetry field looks to be on a scale of 0 to 100 and given that it is an efficiency I would suspect that the units should be percent - please confirm

shambergoldstein (changes requested):
Note: backward incompatible changes in this PR only effect the specific devices that this PR is for.

ontology/yaml/resources/units/units.yaml
evgeniya-engel: It should be kW/ton. I adjusted value range to all flexible between 0 and 10 kw/ton.

shambergoldstein (approved):

PR #1461, opened 2024-11-08
new_SCHWTC_field

kellyhw:
@Ciaran-Oflannagain @kevin-hereworks @cstirdivant


cstirdivant:
@shambergoldstein @ccquigley

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -1408,6 +1408,7 @@ SCHWTC:
   - cooling_percentage_command
   - cooling_request_count
   - cooling_run_command
+  - cooling_run_status
```
shambergoldstein, line 1411: What is the analytical benefit in bringing in this field? You should be able to use the chiller run statuses to infer if cooling is running.

PR #1464, opened 2024-11-12

mschlaffer:
PiperOrigin-RevId: 695777455

Improve default parameters and output for universe selection and UDMI in toolkit and instance_validator.

Details of changes:
- Aligned udmi default parameter in instance_validator to toolkit default: False
- Ensure toolkit correctly uses "default types path" instead of "modified types path" if no ontology changes are set. This ensures instance validator can skip the lengthy validation step when it isn't needed
- Also added some output to make it clear which universe got selected and that validation is running if a modified universe gets selected

### DIFF
--- a/tools/toolkit.py
+++ b/tools/toolkit.py
@@ -71,9 +71,10 @@ def _ParseArgs() -> argparse.ArgumentParser:
       '--modified-ontology-types',
       dest='modified_types_filepath',
       required=False,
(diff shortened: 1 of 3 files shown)


shambergoldstein (approved):

mschlaffer:
tests are failing during python dependency setup, unrelated to this change. needs to be looked at separately. tests passed in google3 before sync and I tested manually from this branch. should be safe, merging.

PR #1485, opened 2024-12-19
Connections Documentation

shambergoldstein:


bsimmons-onboard:
Are there any additional guardrails to consider when modeling these connections? 
For example does FEEDS apply to all entity types including HVAC & FACILITY?

Can HAS_PART apply to facility entity types?

shambergoldstein:
> o all entity types including HVAC & FACILITY?
> 
> Can HAS_PART apply to facility entity

@bsimmons-onboard I added in a bit of guidance for this on some of the connection types, but kept it to only where I felt comfortable. I think it is best to keep it more open overall as I think there could be edge cases for each and I don't want to mislead. Has part for example I think could theoretically include a facility entity; maybe a floor could have part of a wing.

tomfwilliams (approved):
LGTM

bsimmons-onboard:
Thanks for the updates. If the intention is that FEEDs can be used as an equipment FEEDs a FACILITY entity can we add that explicitly to the description?

ccquigley (approved):

PR #1499, opened 2025-01-24
dehumidifier addtions

ccquigley:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -102,7 +102,17 @@ OADM:
   - failed_outside_air_damper_alarm
   uses:
   - outside_air_damper_command
+  - outside_air_damper_status
```
shambergoldstein, line 105: I see a few devices using this abstract, lets move this field to opt_uses so they dont break

shambergoldstein (changes requested):

shambergoldstein (approved):

PR #1503, opened 2025-02-05
Adding Kitchen Hood

darshanjain12:
Adding a Kitchen Hood with Motorized Volume Control Dampers for Fresh Air and Exhaust Air, along with Common Hood Status Monitoring


ontology/yaml/resources/HVAC/entity_types/HOOD.yaml
shambergoldstein, line 1: Is the fresh air damper fed directly from outside air? If so we should use an outside air damper abstract type for it.

shambergoldstein (changes requested):

ontology/yaml/resources/HVAC/entity_types/HOOD.yaml
darshanjain12: @shambergoldstein Fresh Air damper is fed air from a treated fresh air unit (TFA)

shambergoldstein (approved):

PR #1517, opened 2025-02-28
New multi-zone humidity monitoring and AHU type

DaoChen701:


evgeniya-engel:
@shambergoldstein @cquigley2  ready for your review. Thank you!

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
shambergoldstein, line 1: I have concerns with creating abstract types like this as the potential variations in sensor numbers could really blow up the number of abstract types in dbo unnecessarily. Are these sensors used in control at all? Is there an object for the average or max that we could use instead? If answer is no to both of these I think we could probably leave them out of the translation.

shambergoldstein: Do they have alarms for high/low humidity set up using the sensors in the BMS? If not I think we can consider them non-critical and omit them.

DaoChen701: They have high humidity alarms in the BMS.

shambergoldstein: Okay lets keep them then. Are the high humidity alarms bacnet available? If so I would add them as optional on the abstract.

DaoChen701: Yes, they are bacnet available. I will add them as optional.

evgeniya-engel: Yeah they have high alarms.

ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -9293,3 +9293,27 @@ SDVM:
   - supply_air_damper_voltage_sensor
   implements:
   - MONITORING
+
+ZHM5X:
+  guid: "6aa8ece3-5ec0-4c66-9d90-e397bab9d15b"
+  description: "Five sensor zones humidity monitoring"
```
shambergoldstein, line 9299: Please change "zones" to "zone"

shambergoldstein (approved):

PR #1518, opened 2025-03-06
ZOC- adding opt_uses to count people in and out

kellyhw:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
shambergoldstein, line 1: What does count in and count out mean? Is it when someone enters or exits?

kellyhw: Hi @shambergoldstein, 
yes, it is when someone enters or exists and it accumulates.

shambergoldstein: Does any of the equipment control to when someone enters or exits? I am struggling to see the analytical benefit of including these fields.

kellyhw: For data collection, no control. To know how many people enter and exit the room, it does accumulate the values to define the device accuracy.

shambergoldstein: If you feel these are important to include for your use case then you will need to create the fields in order to add them to an abstract type. Please be sure to follow the stucture decribed here: https://github.com/google/digitalbuildings/blob/master/ontology/docs/learning/Module_1_Lesson_4_Fields.pdf in otder for them to pass validation. I would encourage you to use existing subfields as much as possible, i.e. "entering", "leaving", and adjust the subfield definitions as needed.

shambergoldstein: Lets change "count" to "accumulator" on the field names since the number can't decrease as a zone_occupancy_count can. Additionally, please change the definition for the "leaving" subfield to be more broad so it makes sense for this use case. Something similar to the "entering" definition should be good.

shambergoldstein: Please keep the field using "leaving". I meant to please alter the "leaving" subfield definition here: https://github.com/google/digitalbuildings/blob/master/ontology/yaml/resources/subfields/subfields.yaml. "Exiting" does not exist as a defined subfield so validation fails. Sorry for the confusion.

kellyhw: Thank you for the review.

shambergoldstein: So sorry, it looks like "accumulator" actually requires a measurement subfield with units. Lets change them back to count, then we should be good to merge.

shambergoldstein (approved):

PR #1521, opened 2025-03-20
Fixing github actions versions by pinning commit versions.

CauchyPeano:


mschlaffer (approved):
Looks good, please add a description to the PR though. you can reference this as the trigger https://github.com/tj-actions/changed-files/blob/678ecec17288a6a0f240e421c22d5641e0d4e8d7/README.md?plain=1#L18-L36

PR #1536, opened 2025-04-18
Ontology extension on FCU

DaoChen701:


ontology/yaml/resources/HVAC/entity_types/FCU.yaml
shambergoldstein, line 1: I wouldn't expect MTM on an FCU which should not perform any mixing of air - only recirculate return air. Similarly, I am confused how the return air damper would work here as all of the air should be coming from the return. Does the fan speed and cooling adjust with it?

shambergoldstein (approved):

PR #1540, opened 2025-04-22
AHU Update

hhunterr71:
AC is able to run based on schedule or zone occupancy status.

### DIFF
--- a/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
+++ b/ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
@@ -1142,7 +1142,6 @@ SFSS:
   - supply_fan_lost_power_alarm
   - schedule_run_command
   - optimal_stop_command
-  - ventilation_request_count
   uses:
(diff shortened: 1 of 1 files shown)


hhunterr71:
@shambergoldstein @ccquigley Ready for review.

shambergoldstein:
@hhunterr71 I am assuming this is a multi-zone ahu since it is using supply fields so what is the meaning of zone_occupancy_status? Which zone does it refer to?

hhunterr71:
@shambergoldstein This unit is multizone and has a totalizer for occupancy, any one of them can trigger the unit to run.

ccquigley:
Hey @hhunterr71 do you mind confirming in the VAV control programs that these VAVs have occupancy sensors and its not a point name that is misleading? It's pretty common for VAVs and AHUs to have points with occupancy in the name that are defined just using the schedule and no actual occupancy sensor

hhunterr71:
@ccquigley Just removed that occupancy sensor point. Ready for review.

ccquigley (approved):

PR #1547, opened 2025-04-30
new ashp, hws, fcu types and new alarm points

ccquigley:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -7537,6 +7546,7 @@ HWRWISOVM:
   opt_uses:
   - failed_heating_return_water_isolation_valve_alarm
   - heating_return_water_isolation_valve_percentage_sensor
+  - heating_return_water_isolation_valve_percentage_command
```
shambergoldstein, line 7549: How does this work that it has both a percentage command and a binary command? Why not just use HWRWISOVPM?

shambergoldstein (approved):

PR #1550, opened 2025-05-06
AHU Addition

hhunterr71:
One new type and a change to ECON to fit for a new AHU type.


ontology/yaml/resources/HVAC/entity_types/AHU.yaml
```diff
@@ -2597,6 +2597,21 @@ AHU_BSPC_ECON_EFSS_SARC_SFSS_SFVSC_SSPC:
   - SFVSC
   - SSPC
 
+AHU_BSPC_ECON_EFSS_EFVSC_SARC_SFSS_SFVSC_SSPC:
```
shambergoldstein, line 2600: please add an abstract for cooling (dx or chw)

shambergoldstein (changes requested):

PR #1589, opened 2025-06-26
Updates to GENERALTYPES.yaml and telemetry_fields.yaml

eaadalenDB:
Updates to GENERALTYPES.yaml and telemetry_fields.yaml


ontology/yaml/resources/HVAC/entity_types/GENERALTYPES.yaml
shambergoldstein, line 1: Sensor already exists in the global namespace so we do not need to add it here as well. You just need to format your canonical type implements to reference `/SENSOR` with the slash included. Please see <https://github.com/google/digitalbuildings/blob/master/ontology/yaml/resources/HVAC/entity_types/SENSOR.yaml> file for examples

shambergoldstein (changes requested):

PR #1643, opened 2025-10-07
New type for MAU

DaoChen701:


shambersafai (changes requested):
please add a new type instead of changing an existing one

DaoChen701:
The existing one was pushed last week. This is revised for the same MAU.

evgeniya-engel:
@shambersafai confirming that we have revised the MAU type that Dao has added in his previous PR.

shambersafai (approved):

PR #1716, opened 2026-05-01
Optimal Control Type Updates

tasodorff:
Some updated types to cover new OC sites.

Note that boiler is non-canonical due to extra points that are fairly generic but not very common. Instead of polluting abstract type definitions with more optional fields, just adding them directly to the non-canonical type definition.


shambersafai (changes requested):
Can you add guids? Doesn't look like the auto-add worked

shambersafai (approved):

PR #1718, opened 2026-05-05
Ontology updates

eaadalenDB:


ontology/yaml/resources/HVAC/entity_types/ABSTRACT.yaml
```diff
@@ -7914,6 +7916,8 @@ EHHRC:
   - exhaust_air_flowrate_sensor
   - heat_recovery_water_isolation_valve_command
   - leaving_heat_recovery_coil_temperature_sensor
+  opt_uses:
+  - flowrate_sensor
```
shambersafai: I would make this more specific. Im assuming this is the water flowrate since the type already has exhaust_air_flowrate_sensor, so I would make it heat_recovery_water_flowrate_sensor.

ontology/yaml/resources/HVAC/entity_types/FAN.yaml
```diff
@@ -891,6 +891,18 @@ FAN_VSC_EPC_EDM:
   - EPC
   - EDM
 
+FAN_SS_VSC_EPC_EHHRC_EDPM:
```
shambersafai: Just use VSC instead of SS_VSC.

shambersafai (changes requested):

shambersafai (approved):

PR #1724, opened 2026-06-03
More FDD Point Additions

shambersafai:


shambersafai:
Not finalized, do not merge

ontology/yaml/resources/fields/telemetry_fields.yaml
```diff
@@ -893,6 +899,12 @@ literals:
 - discharge_fan_voltage_sensor:
     flexible_min: 0.0
     flexible_max: 500.0
+- discharge_pressure_sensor:
```
eaadalenDB, line 902: What's the use case for this? When would we not want to specify what type of equipment is discharging?

eaadalenDB (commented):
Left a comment

ontology/yaml/resources/fields/telemetry_fields.yaml
shambersafai: Its for pumps. Since pump is inherant in the general type we dont need to specify in the field names.

eaadalenDB (approved):
LGTM

PR #1742, opened 2026-09-18
Edit new fan type

shambersafai:
Update fan type - this is a recently added type and is not used by any devices yet


eaadalenDB (approved):
LGTM

